PR #1, opened 2017-07-17
Query power once per controller, not client.

jhobbs:
This fixes the underlying slowness part of bug 1704489 for me.


blakerouse:
We don't accept pull requests to MAAS on github, but you where down the correct path. But this is a broader problem in MAAS and needs a the change higher up.

https://code.launchpad.net/~blake-rouse/maas/+git/maas/+merge/327585

PR #2, opened 2019-01-21
Support for Nova power type Keystone v3 Auth

degorenko:
Since OpenStack Queens Keystone supports ONLY v3 auth
hence parameters below are required to work with that version
but old one are left as an option for backward compatibility
with prev OpenStack versions.

[0] https://docs.openstack.org/python-novaclient/latest/reference/api/index.html


degorenko:
@blakerouse  since you're initially author for Nova Power type - could you please check this one? Thanks in advance.

PR #6, opened 2019-08-24
Add class typehints to avoid LGTM false positive

cclauss:
Workaround for https://discuss.lgtm.com/t/python-false-positive-non-callable-initiating-class-from-classmethod/2261 ?


blakerouse:
Thank you for your contribution, at the time we don't accept pull requests to MAAS on github, please file the bug and MP on launchpad.net/maas.

PR #8, opened 2019-11-23
Request to supports SSH Power driver

os369510:
In our development environment, we need to daily deploy the daily built image to various platforms. (such as workstation, laptop, IoT board, etc...)
There's many platforms not support either:
1. Power on when AC detected (which means can not be wake up by PDU).
2. Wake on LAN.

and also these platforms won't be power-off after last deployment completed.

Here is the idea to power control device through ssh command.
Requirements:
1. Generate ssh keys on MaaS server (e.g. put ssh key in ```/var/lib/maas/id_rsa```).
2. Add the pair public key of above one to MaaS account.
3. Assign a static IP for target device.
4. Need to know a super user name on target device.

Then the device would be reboot after 5 secs (to prevent got ssh connection closed error) when deploying image if the device has been deployed by MaaS in previous stage.

I've already test it with ```version 2.5.0-7442-gdf68e30a5-0ubuntu1~18.04.1```.
Screenshot from 2019-11-23 16-15-17

I'm new in MaaS world and would like to know more opinions.
Please let me know if anything I'm missing.

### DIFF
--- a/src/provisioningserver/drivers/power/registry.py
+++ b/src/provisioningserver/drivers/power/registry.py
@@ -28,6 +28,7 @@ from provisioningserver.drivers.power.ucsm import UCSMPowerDriver
 from provisioningserver.drivers.power.virsh import VirshPowerDriver
 from provisioningserver.drivers.power.vmware import VMwarePowerDriver
 from provisioningserver.drivers.power.wedge import WedgePowerDriver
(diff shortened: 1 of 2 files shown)


blakerouse:
Thank you for the contribution. All code contributions must go through launchpad.net/maas. Please propose this branch there for discussion. It would also be good to open a feature discussion on the discourse.maas.io

PR #15, opened 2022-07-04
Fix bmc_config.py for moonshot deployment

tata2:
When trying MAAS to deploy an HP moonshot node, it was not working. I fixed it by modifying the bmc_config.py file.

It seemed that some typos were present inside the file.


troyanov:

PR #17, opened 2022-08-23
Pyre: domainname has type `str` but used as type `None`.

luca-digrazia:
**"filename"**: "src/maasserver/api/dnsresources.py"
**"warning_type"**: "Incompatible variable type [9]"
**"warning_message"**: " domainname is declared to have type `str` but is used as type `None`."
**"warning_line"**: 74
**"fix"**: remove `None`


troyanov:

PR #18, opened 2022-08-25
Support for Moonshot Chassis 2.0

forky2:
Patch to provide Moonshot Chassis 2.0 firmware support.

When adding a Moonshot chassis, the driver will correctly ascertain whether firmware is v1 or v2 and send appropriate discovery commands according to version.


troyanov:

PR #137, opened 2026-04-09
fix: change python version used by RTD

AloizioMacedo:
Our code uses some things that are not available in 3.11, like `override` from typing.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -7,7 +7,7 @@ version: 2
 
 # Set the version of Python and other tools you might need
 build:
-  os: ubuntu-22.04
+  os: ubuntu-24.04
   tools:
     python: "3.12"
   jobs:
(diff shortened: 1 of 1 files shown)


MaikRe (dismissed):
+0.01

.readthedocs.yaml
```diff
@@ -9,7 +9,7 @@ version: 2
 build:
   os: ubuntu-22.04
```
alexsander-souza: why not noble?

AloizioMacedo: This was mainly preserving the same [settings found on the starter pack](https://github.com/canonical/sphinx-docs-starter-pack/blob/main/.readthedocs.yaml) by default. 

I will change to noble.

MaikRe (approved):
+1

PR #138, opened 2026-04-10
fix: use clean authoritative flag before checking it

AloizioMacedo:
The main issue was that the validation was being done before any treatment of the inputs, so `self.data.get("authoritative")` was returning the string that was used in the CLI command (or the API). 

Specifically, for example, setting `authoritative=false` resulted in the code thinking that there was a violation.

Funnily enough, setting `authoritative` to an empty string would let things pass through, even though authoritative at the end (in the model) would be True due to its default value there.

Resolves [LP:2147570](https://bugs.launchpad.net/maas/+bug/2147570)

### DIFF
--- a/src/maasserver/forms/domain.py
+++ b/src/maasserver/forms/domain.py
@@ -40,13 +40,22 @@ class DomainForm(MAASModelForm):
         return self.instance
 
     def clean(self):
-        if self.data.get("authoritative") and len(
-            self.data.get("forward_dns_servers", "")
+        # Using "cleaned_data" here, inside a `clean` method
+        # that later calls super().clean() looks like things are out
+        # of order, but the way Django works is that this method is
+        # supposed to run _after_ field-level validation has run.
+        #
+        # You can see that the method that this overrides in the parent
+        # specifically says that it is a "hook for doing any extra
+        # form-wide cleaning after Field.clean() has
+        # been called on every field."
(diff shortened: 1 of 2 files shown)


src/maasserver/forms/domain.py
```diff
@@ -40,13 +40,21 @@ def save(self):
...
+        cleaned_data = super().clean()
+
+        authoritative = cleaned_data.get("authoritative")
+        authoritative = (
+            authoritative
+            if authoritative is not None
+            else self.instance.authoritative
+        )
+
+        if authoritative and cleaned_data.get("forward_dns_servers"):
```
wyattrees, line 52:
```
if (authoritative or self.instance.authoritative) and cleaned_data.get("forward_dns_servers"):
````
is a bit cleaner IMO

wyattrees (approved):
one suggestion in-line, otherwise lgtm

src/maasserver/forms/domain.py
AloizioMacedo: `(authoritative or self.instance.authoritative)` would always be `True` on creation of a new Domain, because `self.instance.authoritative` collapses to the default value (which is `True`) in that case.

In the way it is implemented, if someone sets `authoritative=False`, then `authoritative` continues properly as `False` (since authoritative will not be None in that case).

PR #139, opened 2026-04-10
feat(maasapiserver): add /groups:statistics endpoint and allow filtering by name

nehjoshi:
## Done
- Added a statistics endpoint for usergroups to get the `user_count` for each group
- Added a filter so that we can search for a group by its name

Partially resolves [MAASENG-6347](https://warthogs.atlassian.net/browse/MAASENG-6347)

## Notes
- This PR will allow the creation of the following UI view (note: this screenshot is with actual backend integration and not just sample data)

image

[MAASENG-6347]: https://warthogs.atlassian.net/browse/MAASENG-6347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/maasapiserver/v3/api/public/handlers/usergroups.py
```diff
@@ -67,6 +71,22 @@ class UserGroupsHandler(Handler):
 
     TAGS = ["UserGroups"]
 
+    def get_handlers(self):
```
r00ta: why do you need to override the `get_handlers`?

r00ta (changes requested):
question inline, plus the test case coverage should be improved

src/maasapiserver/v3/api/public/handlers/usergroups.py
nehjoshi: Because without it, it defaults to alphabetical sorting of handlers, so the `/groups/statistics` endpoint always returns a `422: Unprocessable entity` as it's trying to match the `/groups/{group_id}` endpoint.

r00ta: In the API we are trying to follow the policy where the endpoints are structured like `/collection/item/collection` . `/groups/statistics` is not compliant. I'd rather use `/collection:statistics`

alemar99: @r00ta we agreed on the `/<entity>/statistics` patterns in the spec "MA260 - Nested data in the V3 API". That's what we did also on the images

r00ta: If we did a mistake in the MA260 approving something that is not compliant with the API design it does not mean we have to keep going with it.

nehjoshi: To add to what @alemar99  mentioned, `/<entity>/statistics` is also currently the standard in the [MAAS API Guidelines doc](https://docs.google.com/document/d/1eHW0ywXG-tvlGsYNzjk93NVf6N_QP9E8kXQ27rSuDsE/edit?tab=t.0).

That being said, I'm more than happy to change the code if we want to switch to `:statistics`, it's just a one-liner. If we do move forward with` /<entity>:statistics`, I can update the guidelines doc as well.

r00ta: well, I guess somebody added that when they approved the spec. It was definitely not me

nehjoshi: Changed to `:statistics`, I will update the API guidelines too.

src/maasapiserver/v3/api/public/handlers/usergroups.py
```diff
@@ -67,6 +71,22 @@ class UserGroupsHandler(Handler):
...
+        return [
```
r00ta: you can now remove this

r00ta (dismissed):

alemar99 (approved):

PR #140, opened 2026-04-13
refactor: make the boot images import a Temporal schedule with configurable interval

alemar99:


src/maascommon/path.py
```diff
@@ -9,11 +9,3 @@ def get_maas_data_path(path: str) -> str:
     """Return a path under the MAAS data path."""
     base_path = Path(getenv("MAAS_DATA", "/var/lib/maas"))
     return str(base_path / path)
-
-
-def get_maas_lock_path() -> Path:
```
alemar99, line 14: This is not used anymore

src/maasserver/bootresources.py
```diff
@@ -95,70 +86,6 @@ def stop_import_resources():
...
-class ImportResourcesService(TimerService):
-    """Service to periodically import boot resources.
-
-    This will run immediately when it's started, then once again every hour,
-    though the interval can be overridden by passing it to the constructor.
-    """
-
-    def __init__(self, interval=IMPORT_RESOURCES_SERVICE_PERIOD):
-        super().__init__(interval.total_seconds(), self.maybe_import_resources)
-        for p in [get_maas_lock_path(), get_bootresource_store_path()]:
```
alemar99, line 107: This is not needed as these directories are created automatically both in the snap and the deb. See snap/hooks/install and debian/maas-region-api.dirs

src/maastemporalworker/schedules.py
```diff
@@ -29,10 +32,63 @@
...
+        spec=ScheduleSpec(
+            # Will be updated at startup with the value from the db
+            intervals=[ScheduleIntervalSpec(every=timedelta(minutes=60))]
+        ),
+        policy=SchedulePolicy(overlap=ScheduleOverlapPolicy.CANCEL_OTHER),
+    ),
 }
 
 
+def _master_image_sync_updater(sync_interval_minutes: int):
```
alemar99: This was done to be able to test it in an easier way, instead of defining it as a closure

src/maastemporalworker/temporal_script.py
```diff
@@ -273,6 +282,26 @@ async def main() -> None:
...
+        async with conn.begin():
+            services = await ServiceCollectionV3.produce(
+                context=Context(connection=conn), cache=services_cache
+            )
+            configs = await services.configurations.get_many(
+                {
+                    BootImagesImportIntervalMinutesConfig.name,
+                    BootImagesAutoImportConfig.name,
+                }
+            )
```
alemar99, line 294: This is actually the part that I don't really like, but I couldn't think on a better way to retrieve those values at startup

src/maastemporalworker/temporal_script.py
```diff
@@ -273,6 +282,26 @@ async def main() -> None:
...
+
+    await pause_or_unpause_master_image_sync_schedule(
```
r00ta: is this going to actually start the schedule and pause it immediately after?

r00ta (commented):
question inline

src/maastemporalworker/temporal_script.py
alemar99: That's actually a good point. In a normal scenario no, because at first MAAS startup the config value will be True and it will start. On subsequent restarts, since the schedule state changes whenever the config is changed, is not possible to have the schedule active and the config value set to False.

But in general I think that it's better to change the definition of the schedule to be paused by default, and then activated after reading the config. In this way we can also have an hacky way to disable the automatic import of images at first startup (until we have a proper configuration file)

r00ta: Is it possible to create the schedule as disabled by default?

alemar99: yes I've done some tests and it's possible

src/maastemporalworker/schedules.py
```diff
@@ -29,10 +33,74 @@
...
+                    every=timedelta(minutes=sync_interval_minutes)
+                )
+            ]
+        )
+        schedule_description = input.description
+        # When updating a schedule ALL the options must be specified again.
+        # Here we save the current state in order to avoid un-pausing the schedule
+        # when only changing the sync interval
+        current_state = schedule_description.schedule.state
+        if auto_import_enabled_config is not None:
```
r00ta, line 71: I see the config might be theoretically `None`, but maybe we should just remove the `Optional` as it should be always `True` or `False`

alemar99: It's `None` when we update the sync_interval config: https://github.com/canonical/maas/pull/140/changes#diff-113dc369e9d8e9fcf9f90880c70a73cbdbc520ea2866a2c17970c9d23e25d086R70


Also I have to change the name since that method updates the schedule in general, not only the interval

src/maasservicelayer/services/hooked_configurations.py
```diff
@@ -54,6 +60,14 @@ async def set(self, name: str, value: Any) -> None:
...
 
         match name:
+            case BootImagesAutoImportConfig.name:
+                client = await self.temporal_service.get_temporal_client()
+                await pause_or_unpause_master_image_sync_schedule(
+                    client, value
+                )
+            case BootImagesImportIntervalMinutesConfig.name:
+                client = await self.temporal_service.get_temporal_client()
+                await update_master_image_sync_schedule_interval(client, value)
```
r00ta: `value` can be `None` but the signatures accept `int`

alemar99: right.. then I should change the config model definition to not be `Optional`

r00ta (approved):

PR #141, opened 2026-04-13
fix(storage): FilesystemGroup name generation with multiple nodes

alexsander-souza:
FilesystemGroup.get_available_name_for() incorrectly inferred the node when assigning sequential names (md0, md1, bcache0, etc.) to virtual block devices. It used a deprecated Django pattern where unsaved instances passed to queryset filters would accidentally match any node's boot filesystem via WHERE filesystem_group_id IS NULL. In multi-node deployments, this could return an arbitrary node, causing name collisions or IntegrityError on the BlockDevice unique_together constraint.

Resolves LP:2148204

### DIFF
--- a/src/maasserver/models/filesystemgroup.py
+++ b/src/maasserver/models/filesystemgroup.py
@@ -145,16 +145,6 @@ class BaseFilesystemGroupManager(Manager):
                 idx = max(idx, name_idx)
         return f"{prefix}{idx + 1}"
 
-    def get_available_name_for(self, filesystem_group):
-        """Return an available name that can be used for a `VirtualBlockDevice`
-        based on the `filesystem_group`'s group_type and other block devices
-        on the node.
-        """
-        return self.get_available_name_for_node(
-            filesystem_group.group_type,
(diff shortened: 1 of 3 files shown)


src/maasserver/models/filesystemgroup.py
```diff
@@ -151,6 +145,16 @@ def get_available_name_for(self, filesystem_group):
...
+    def get_available_name_for(self, filesystem_group):
+        """Return an available name that can be used for a `VirtualBlockDevice`
+        based on the `filesystem_group`'s group_type and other block devices
+        on the node.
+        """
+        return self.get_available_name_for_node(
+            filesystem_group.group_type,
+            filesystem_group.get_node(),
+        )
+
```
AloizioMacedo, line 148: Do we need to keep this method? I think it is only used in the tests, and if it is used out of them, it could result in the same issue that this PR fixes.

alexsander-souza: removed the method

AloizioMacedo (approved):

PR #154, opened 2026-04-14
docs: replace previous docs with sphinx docs

AloizioMacedo:
This PR replaces the previous documentation with the new one, which uses sphinx and leverages read the docs as hosting platform. As part of this, it does the last step of replacing the offline docs (it is convenient to do both at once when changing `snapcraft.yaml`, since we will not need to temporarily build two docs on the snap).

This will also finalize the transition process.


wyattrees (dismissed):
+1, although it looks like some tests need to either be removed or refactored.

Also, if I'm not mistaken, the description of the PR gets included in the commit message when merge happens, so make sure that it complies with our conventional commits spec

AloizioMacedo:
Putting back to draft due to issues when building the deb. Specifically, there are a number of extensions that use git, and while building the deb we cannot use them.

alexsander-souza (approved):
+1

PR #158, opened 2026-04-15
fix(deps): add assume_scheme to URLField for Django 6.0 compatibility

alexsander-souza:
Django 6.0 will change the default URL scheme from 'http' to 'https'. Explicitly set assume_scheme='http' for URLField declarations to silence RemovedInDjango60Warning and maintain current behavior.

### DIFF
--- a/src/maasserver/fields.py
+++ b/src/maasserver/fields.py
@@ -569,6 +569,10 @@ class URLOrPPAFormField(forms.URLField):
     }
     default_validators = [URLOrPPAValidator()]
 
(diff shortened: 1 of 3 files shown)


alemar99:
Did you miss the two in `src/maasserver/fields.py`?

alexsander-souza:
> Did you miss the two in `src/maasserver/fields.py`?

good catch, I missed one (the other URLField is from another module)

alemar99 (approved):

PR #159, opened 2026-04-15
build: fix binary source paths in debian install files

alexsander-souza:
Corrects the source paths for maas-agent, maas-netmon, and OpenFGA binaries in their respective .install files from bin/ to usr/sbin/, matching the actual build output locations.

### DIFF
--- a/debian/maas-agent.install
+++ b/debian/maas-agent.install
@@ -1,3 +1,3 @@
 debian/extras/99-maas-agent-sudoers etc/sudoers.d
 
-bin/maas-agent usr/sbin
+usr/sbin/maas-agent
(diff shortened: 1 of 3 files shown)


alemar99:
I'm a bit confused. Looking at https://www.debian.org/doc/manuals/maint-guide/dother.en.html#install the first part is where the binary is located and the second part where it should be installed, so they look okay to me.

The built binaries are not inside `bin/`?

The changes that you are proposing don't do basically anything since the binaries are already in the right path?

alexsander-souza:
@alemar99 we do need the changes, but they can be simplified as the source and destination relative paths are the same (just did this). 

We must enumerate the files even if they are already installed in their final destinations because we are building multiple binary packages from one source, so debhelper needs this to know which file to put in each package.

alemar99 (approved):
Thanks for the explanation

PR #160, opened 2026-04-15
chore(packaging): address lintian warnings

alexsander-souza:


troyanov (approved):
LGTM, but I am personally waiting for `deb` support being dropped 😊

PR #163, opened 2026-04-17
feat(maasservicelayer): add bulk-remove and bulk-create operations for users and entitlements in groups

nehjoshi:
Added 3 new endpoints:
1. `POST /groups/{group_id}/members:batch_create`: For adding multiple users to a group in one operation. The `batch_create` suffix is appended to separate this endpoint from the `POST /groups/{group_id}/members` for adding a single member.
2. `DELETE /groups/{group_id}/members`: For removing multiple users from a group in one operation
3. `POST /groups/{group_id}/entitlements:batch_delete`: Removes multiple entitlements from a group in a single operation. Entitlements within a group do not have a unique `id`. Instead, each entitlement is identified by a combination of fields: `resource_type`, `resource_id`, and `entitlement` (e.g. `can_edit_machines`). Deleting multiple entitlements therefore requires sending a structured payload (e.g. an array of these field combinations).
While this could map to a `DELETE` endpoint, doing so would require a request body. In practice, request bodies on `DELETE` are discouraged by MDN (see [docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/DELETE)). For this reason, `POST` is chosen. Also, the suffix `batch_delete` is appended because `POST /groups/{group_id}/entitlements` already exists for entitlement creation.

The following endpoints are modified and now include pagination:
1. `GET /groups/{group_id}/members`
2. `GET /groups/{group_id}/entitlements`

Resolves [MAASENG-6347](https://warthogs.atlassian.net/browse/MAASENG-6347)

[MAASENG-6347]: https://warthogs.atlassian.net/browse/MAASENG-6347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/maasapiserver/v3/api/public/handlers/usergroups.py
+++ b/src/maasapiserver/v3/api/public/handlers/usergroups.py
@@ -1,10 +1,10 @@
 # Copyright 2026 Canonical Ltd.  This software is licensed under the
 # GNU Affero General Public License version 3 (see the file LICENSE).
 
-from typing import Union
+from typing import Annotated, Union
 
 from fastapi import Depends, Header, Query, Response
-from pydantic import conlist
+from pydantic import Field
 from starlette import status
 
 from maasapiserver.common.api.base import Handler, handler
@@ -61,6 +61,7 @@ from maasservicelayer.exceptions.constants import (
     INVALID_ARGUMENT_VIOLATION_TYPE,
     USER_ALREADY_IN_GROUP,
 )
+from maasservicelayer.models.fields import UniqueList
 from maasservicelayer.services import ServiceCollectionV3
 from maasservicelayer.services.openfga_tuples import EntitlementsBuilderFactory
 from maasservicelayer.services.usergroups import (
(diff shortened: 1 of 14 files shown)


src/maasapiserver/v3/api/public/handlers/usergroups.py
```diff
@@ -556,3 +677,56 @@ async def remove_group_entitlement(
             group_id, entitlement, resource_type, resource_id
         )
         return Response(status_code=status.HTTP_204_NO_CONTENT)
+
+    @handler(
+        path="/groups/{group_id}/entitlements:batchDelete",
```
r00ta: why camel case?

nehjoshi: Good catch, this was an oversight. I have changed them to snake_case

src/maasservicelayer/services/openfga_tuples.py
```diff
@@ -266,3 +311,24 @@ async def delete_entitlement(
             )
         )
         await self.delete_many(query)
+
+    async def bulk_delete_entitlements(
+        self,
+        group_id: int,
+        items: list[tuple[str, str, int]],
```
r00ta: In general I discourage to pass `list[tuple[str,str,int]]`. Use a typed model instead

r00ta (changes requested):

src/maasservicelayer/services/openfga_tuples.py
nehjoshi: Noted, I have changed this to an `EntitlementDeleteSpec` model

r00ta (approved):
lgtm

PR #166, opened 2026-04-17
feat: use correct MAAS logo

MaikRe:
Adds icons from assets.ubuntu.com


MaikRe:
~Pulled from the assets folder in our drive~

AloizioMacedo (changes requested):
Thanks, Maik!

- I think you have to uncomment L215 (specifying the static path). I had to, when testing locally.
- Can you run the image through the `compress.sh` script? Should shave some kbs.

MaikRe:
> Thanks, Maik!
> 
> * I think you have to uncomment L215 (specifying the static path). I had to, when testing locally.
> * Can you run the image through the `compress.sh` script? Should shave some kbs.

I have uncommented the static path.
I do not want to run the image through the compression script as that would modify it from the Canonical original.

AloizioMacedo (approved):

PR #167, opened 2026-04-17
fix(docs): fix docs deps after resolute work

AloizioMacedo:
This is the low-effort fix. Since the docs build is a separate process, this should be enough.

In the long term, ideally we want to simply install MAAS as a package when building the documentation, and the granular deps in the docs to be pertaining only to the docs itself.

### DIFF
--- a/rtd-docs/requirements.txt
+++ b/rtd-docs/requirements.txt
@@ -26,7 +26,7 @@ rst2html==2020.7.4
 vale==3.13.0.0
 
 # Bare minimum MAAS dependencies for API documentation generation
-django==6.0.4
+django==5.2.9
 git+https://git.launchpad.net/django-piston3
 aiofiles==25.1.0
 aiohttp==3.13.5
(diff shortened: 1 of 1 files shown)


rtd-docs/requirements.txt
```diff
@@ -26,7 +26,7 @@ rst2html==2020.7.4
 vale==3.13.0.0
 
 # Bare minimum MAAS dependencies for API documentation generation
-django==4.2.11
+django==6.0.4
```
alexsander-souza: running 5.2.9 on Resolute

rtd-docs/requirements.txt
```diff
@@ -48,7 +49,7 @@ petname==2.9
 pexpect==4.9.0
 prometheus_client==0.24.1
 psycopg2-binary==2.9.11
-pydantic==1.10.26
+pydantic==2.13.2
```
alexsander-souza: running 2.12.5

alexsander-souza (changes requested):
the django version is going to break things

AloizioMacedo:
> the django version is going to break things

I changed things to be in line with what you specified to be extra safe.

But worth noticing that these discrepancies are fine (as long as the docs build): the docs building environment is isolated from the code. We just need the "critical" paths to be consistent in order for the code that gathers the OpenAPI spec to work without hitting issues.

alexsander-souza (approved):
+1

PR #168, opened 2026-04-20
refactor: handle boot source url updates in API v2

alemar99:
In v2, the url of a boot source can be changed. This can result in inconsistencies between the selection already present and the selection in the (possibly) new stream.
To handle this, a new workflow is triggered when the URL is being changed. It will:
- update the boot source cache
- create a warning notification for the selections that aren't present anymore
- synchronize the selections that are still available AND with the highest priority


alexsander-souza (approved):
+1

PR #169, opened 2026-04-20
fix: do not fail deploying node actions when deploy wf no longer exist

skatsaounis:
- Properly check for no workflows found error (`RPCError` with status `RPCStatusCode.NOT_FOUND`)
- Remove `WorkflowQueryFailedError` blocks as `cancel()` is not throwing these errors at all and it was dead code

Resolves: LP:2148683

### DIFF
--- a/src/maasserver/models/node.py
+++ b/src/maasserver/models/node.py
@@ -62,7 +62,7 @@ from django.shortcuts import get_object_or_404
 from django.utils import timezone
 from netaddr import IPAddress, IPNetwork
 import petname
-from temporalio.client import WorkflowFailureError, WorkflowQueryFailedError
+from temporalio.client import WorkflowFailureError
 from temporalio.common import WorkflowIDReusePolicy
 from twisted.internet import reactor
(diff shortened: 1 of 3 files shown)


skatsaounis:
**NOTES:**

- https://python.temporal.io/temporalio.client.WorkflowHandle.html#cancel reference to cancel and the only error that throws `RPCError`
- https://github.com/temporalio/temporal/issues/4118 there is no documentation around it but this issue and the fix reveal what server sets in the RPCError status. The message is changing between versions (after the fix) but the stats remains the same.
- We have other two functions that are not properly handling cancel but this PR tries to address only the linked bug
  - https://github.com/canonical/maas/blob/11de5490bf650b3707b7bf1484cb5bba3b055297/src/maasserver/workflow/__init__.py#L129
  - https://github.com/canonical/maas/blob/11de5490bf650b3707b7bf1484cb5bba3b055297/src/maasserver/workflow/__init__.py#L116

alemar99 (approved):
LGTM

PR #170, opened 2026-04-20
fix(ci): added explicit core26 install to maas-ui-sync.yaml

abuyukyi101198:
## Done

- Added explicit `core26` install step to support resolute images


tmerten (approved):
that should work

PR #171, opened 2026-04-20
fix: minimum DPU kernel validation

motrilense:
This PR fixes a DPU deploy validation bug for custom images where MAAS could reject a valid kernel.

Fix
- Correct the DPU kernel comparison logic so it rejects only kernels older than `min_hwe_kernel`
- Keep the existing minimum floor validation for DPU (hwe-22.04)
- Preserve behavior for truly invalid older kernels

Add regression tests
- test_dpu_hwe_kernel_equal_to_min_hwe_kernel_is_valid
- test_dpu_hwe_kernel_older_than_min_hwe_kernel_is_invalid

Resolves [LP:2148598](https://bugs.launchpad.net/maas/+bug/2148598)


alexsander-souza (approved):
+1

PR #173, opened 2026-04-21
fix(ci): use edge channel for core26

abuyukyi101198:
## Done

- Added `--edge` flag for `core26` snap install


nehjoshi (approved):
Lgtm. I tried setting it up with `--edge core26` yesterday and it has worked, so this looks good.

PR #165, opened 2026-04-17
chore(ci): pick an image when manually running

alexsander-souza:
Always use the target branch as the image tag: merge_group.base_ref
for merge queue runs, base_ref for PR runs, and ref_name for manual
runs (workflow_dispatch).

Compute the tag once in the changes job and expose it as an output
to avoid repeating the logic in every container job.

PR #176, opened 2026-04-21
fix: correct merge-group image tag

motrilense:
This PR fixes a bug in the CI pipeline:
- Docker pull fails on `merge_group` runs: `github.event.merge_group.base_ref` returns a full ref (refs/heads/master) instead of a short branch name, producing an invalid Docker image tag. Strip the refs/heads/ prefix with bash parameter expansion so the tag is consistent across all trigger types.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -31,7 +31,9 @@ jobs:
 
       - name: Set build env image tag
         id: image
-        run: echo "tag=${{ github.event.merge_group.base_ref || github.base_ref || github.ref_name }}" >> $GITHUB_OUTPUT
+        run: |
+          ref="${{ github.event.merge_group.base_ref || github.base_ref || github.ref_name }}"
+          echo "tag=${ref#refs/heads/}" >> $GITHUB_OUTPUT
 
       - uses: dorny/paths-filter@61f87a10cd2c304679af17bb73ef192addf33c1c
         id: filter
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -152,5 +152,5 @@ jobs:
     needs: [lint, test-python, test-go, test-docs]
     steps:
       - name: Check job results
-        if: needs.*.result == 'failure'
+        if: contains(needs.*.result, 'failure')
         run: exit 1
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -154,5 +154,5 @@ jobs:
     needs: [lint, test-python, test-go, test-docs]
     steps:
       - name: Check job results
-        if: contains(needs.*.result, 'failure')
+        if: needs.*.result == 'failure'
         run: exit 1
--- a/src/maasserver/forms/__init__.py
+++ b/src/maasserver/forms/__init__.py
@@ -927,7 +927,9 @@ class MachineForm(NodeForm):
             if not min_hwe_kernel:
                 cleaned_data["min_hwe_kernel"] = min_hwe_kernel = "hwe-22.04"
             elif not release_a_newer_than_b(min_hwe_kernel, "hwe-22.04"):
-                raise ValidationError(f"Invalid minimum {min_hwe_kernel} DPU kernel used. Only 'hwe-22.04' or newer are supported.")
+                raise ValidationError(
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -152,5 +152,5 @@ jobs:
     needs: [lint, test-python, test-go, test-docs]
     steps:
       - name: Check job results
-        if: needs.*.result == 'failure'
+        if: contains(needs.*.result, 'failure')
```
troyanov: Nice catch 👍

troyanov (dismissed):

.github/workflows/ci.yaml
```diff
@@ -31,7 +31,9 @@ jobs:
 
       - name: Set build env image tag
```
troyanov, line 32: I am not sure if your changes are fixing this completely. 
I left a comment at https://github.com/canonical/maas/pull/165#discussion_r3113611425

motrilense: Ah, nice to see that this was already spotted.

For what I could see the docker image can be pulled when a PR runs the CI, but fail to pull when the CI runs the after merged is successful.

In a PR [link -> initialize containers > starting job container](https://github.com/canonical/maas/actions/runs/24680665479/job/72176760514) is pulling the docker image from something like this
`/usr/bin/docker --config /home/runner/work/_temp/.docker_c8203d57-0a33-4a85-b120-38558b6efbd1 pull ghcr.io/canonical/maas-build-env:master`

But once merged, [link -> initialize containers > starting job container](https://github.com/canonical/maas/actions/runs/24709512124/job/72270471042) is pulling it from
`/usr/bin/docker --config /home/runner/work/_temp/.docker_39870636-e107-421a-9e35-8d3440216e37 pull ghcr.io/canonical/maas-build-env:refs/heads/master`

I don't even think that is a valid tag. So I just stripped the prefix "refs/heads/" in case that is present

troyanov: Original workflow definition had this:
```
image: ghcr.io/${{ github.repository }}-build-env:${{ github.event_name == 'merge_group' && 'master' || github.base_ref }}
```

Maybe this is a more robust way just to take `master` for any `merge_group` event (considering that we have this `merge_group` only for a master branch.

What do you think?

motrilense: If we are sure we want to keep using `master`, it might be the safer option. I am just afraid that if we merge into a release branch or something similar (a feature branch maybe?), hardcoding `master` might become an issue. We would have to revert the changes, and being back at this same point

troyanov: It is only a master branch where we want merges to be performed via a merge queue (because of a high pace of PRs). Our versioned branches cannot have merge queue policy

troyanov: Actually, I agree with you, having a more dynamic solution for future branches would be a more safe approach.

.github/workflows/ci.yaml
```diff
@@ -31,7 +31,9 @@ jobs:
...
         id: image
```
troyanov, line 33: @motrilense maybe as a part of this work we can improve naming a bit:
```
image: ghcr.io/${{ github.repository }}-build-env:${{ needs.changes.outputs.image }}
```

I really don't like that we have `image` being actually an `image-tag`
Maybe instead we can have a complete image name being crafted in that step, or just rename variable to `image-tag`. What do you think?

.github/workflows/ci.yaml
bjornt: Could you land this as a separate commit, before you fix the lint job? From what I understand from the workflow runs, this code is not even being run. Is it?

motrilense: Sure, I will prepare another PR and provide some more information there

troyanov (approved):

.github/workflows/ci.yaml
motrilense: https://github.com/canonical/maas/pull/178

.github/workflows/ci.yaml
motrilense: For the time being, we are going to leave it as it is and test the CI fix :)
I am resolving conversation

PR #177, opened 2026-04-21
feat: add full support for multiple boot sources

alemar99:
Full support of multiple boot sources both in the v2 and v3 API. The decision between which selection should be downloaded on which boot source is done based on the boot source's priority (higher is better).

Summary of what changed:
* MAAS will always have the stable and candidate stream boot sources available.
They are immutable, meaning that they can't be deleted nor modified, except for the `priority` and `enabled` fields. The priorities will be by default 1 for candidate and 2 for stable.
* Boot sources now have an `enabled` field: an enabled boot source works as usual, while a disabled boot source is one that will be ignored when synchronizing images. A disabled boot source can still have its selections and its downloaded boot source but as long as it's disabled, no updates will be downloaded.
* Possibility to delete a selection for the latest Ubuntu LTS: since now we have multiple boot sources, the user might want to delete it from one and add it to another. A notification will be raised in case the selection is not present as always.
* A warning will be presented to the user in case all of the boot sources are disabled
* No automatic creation of boot sources or selections at runtime: the user will be responsible for managing them and we'll not do anything smart.
* Drop of the env variable `MAAS_DEFAULT_IMAGES_URL` to override the default boot source for MAAS: it wasn't documented anywhere and likely never used.
* Additions to the v2 API to support these changes: the boot sources v2 API now have the `priority`, `name`, `enabled`, `skip_keyring_verification` **optional** arguments. This will not break the API contract and will be compatible to what we have in v3


alemar99:
Diff is really large mainly because of the re-generation of the initial SQL file.

src/maasservicelayer/db/alembic/versions/0005_add_columns_to_bootsource.py
```diff
@@ -9,63 +9,231 @@
...
+    op.get_bind().execute(
+        text("""
+            INSERT INTO maasserver_bootsourceselection (created, updated, os, release, arches, subarches, labels, boot_source_id)
+            VALUES (:created, :updated, :os, :release, :arches, '{*}', '{*}', :boot_source_id)
+        """),
+        {
+            "created": now,
+            "updated": now,
+            "os": "ubuntu",
+            "release": "noble",
```
alexsander-souza: move this to a constant, making easier to spot what needs to be updated when Resolute becomes the new default

src/tests/maasservicelayer/db/repositories/test_domains.py
```diff
@@ -70,19 +70,9 @@ def repository_instance(
     async def _setup_test_list(
         self, fixture: Fixture, num_objects: int
     ) -> Sequence[Domain]:
-        # The default domain is created by the migrations
-        # and it has the following timestamp hardcoded in the test sql dump,
-        # see src/maasserver/testing/inital.maas_test.sql:8977
-        ts = datetime(2025, 10, 17, 10, 15, 20, 698940, tzinfo=timezone.utc)
+        # The default domain is in the migrations
```
alexsander-souza, line 73: scope creep?

src/tests/maasservicelayer/db/repositories/test_package_repositories.py
```diff
@@ -41,43 +37,10 @@ def repository_instance(
     async def _setup_test_list(
         self, fixture: Fixture, num_objects: int
     ) -> list[PackageRepository]:
-        # The default package repositories are created by the migration and it
-        # has the following timestamp hardcoded in the test sql dump,
-        # see src/maasserver/testing/inital.maas_test.sql:9243
-        ts = datetime(2025, 10, 17, 10, 15, 20, 698940, tzinfo=timezone.utc)
+        # The default package repositories are created by the migrations
```
alexsander-souza, line 40: scope creep?

src/tests/maasservicelayer/db/repositories/test_resource_pools.py
```diff
@@ -57,17 +57,9 @@ async def _setup_test_list(
...
-        ts = datetime(2025, 10, 17, 10, 15, 20, 698940, tzinfo=timezone.utc)
         created_resource_pools = [
-            ResourcePool(
-                id=0,
-                name="default",
-                description="Default pool",
-                created=ts,
-                updated=ts,
-            )
+            ResourcePool(**row)
```
alexsander-souza, line 61: same

src/tests/maasservicelayer/db/repositories/test_usergroups.py
```diff
@@ -67,30 +67,9 @@ def repository_instance(
     async def _setup_test_list(
         self, fixture: Fixture, num_objects: int
     ) -> list[UserGroup]:
-        # The default groups are created by the migrations, they have the following timestamp hardcoded in the test sql dump
+        # The default groups are created by the migrations
```
alexsander-souza, line 70: same

src/tests/maasservicelayer/db/repositories/test_zones.py
```diff
@@ -57,18 +56,9 @@ def repository_instance(
     async def _setup_test_list(
         self, fixture: Fixture, num_objects: int
     ) -> list[Zone]:
-        # The default zone is created by the migration and it has the following
-        # timestamp hardcoded in the test sql dump,
-        # see src/maasserver/testing/inital.maas_test.sql:9558
-        ts = datetime(2025, 10, 17, 10, 15, 20, 698940, tzinfo=timezone.utc)
+        # The default zone is created by the migrations
```
alexsander-souza, line 59: same

alexsander-souza (commented):
some comments inline

src/tests/maasservicelayer/db/repositories/test_domains.py
alemar99: I had to re-generate the initial SQL file and so timestamps changed. Using the fixture now is a more robust way of handling default objects in the migrations

src/maasserver/models/signals/bootsources.py
```diff
@@ -13,23 +13,10 @@
...
-
-def delete_boot_source(sender, instance, *args, **kwargs):
-    """Wrap update_boot_source_cache and ignore the signal arguments"""
-    update_boot_source_cache()
+    update_boot_source_cache(instance.id)
 
 
 signals.watch(post_save, save_boot_source_cache, BootSource)
-signals.watch(post_delete, delete_boot_source, BootSource)
-
```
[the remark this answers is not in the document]
alemar99: deleting a boot source deletes its boot source cache

alemar99:
@alexsander-souza re-review?

src/maasserver/models/bootsourceselection.py
```diff
@@ -186,21 +166,3 @@ def force_delete(self):
         )
         boot_resources_to_delete.delete()
         return super().delete()
```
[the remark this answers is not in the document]
alexsander-souza: update the method signature

alexsander-souza (dismissed):
+1 with one comment inline

alexsander-souza (dismissed):
Image

alexsander-souza (dismissed):

motrilense (approved):
LGTM, approval was reset due to the merge conflict fix

PR #179, opened 2026-04-21
feat: send version in Site Manager heartbeats, known config options when requested

wyattrees:
Updates the Site Manager heartbeat workflow to send the current MAAS version and known configuration options when requested with each heartbeat. Until the configuration workflow is implemented, an empty list is sent as the known configuration options.

### DIFF
--- a/src/tests/maastemporalworker/workflow/test_msm.py
+++ b/src/tests/maastemporalworker/workflow/test_msm.py
@@ -690,7 +690,9 @@ class TestMSMHeartbeatWorkflow:
             input: MSMHeartbeatParam,
         ) -> tuple[int, bool]:
             calls["msm-send-heartbeat"].append(True)
-            return (-1, False)
+            if len(calls["msm-send-heartbeat"]) == 2:
+                return (-1, False)
+            return (1, True)
 
         @activity.defn(name=MSM_GET_ENROL_ACTIVITY_NAME)
         async def get_enrol() -> dict[str, Any]:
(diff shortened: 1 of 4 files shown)


src/tests/maastemporalworker/workflow/test_msm.py
```diff
@@ -497,6 +510,21 @@ async def test_restore_default_bootsource_activity(
...
+        assert config_opts == []
+
+    async def test_get_running_version(self, msm_act):
+        env = ActivityEnvironment()
+        version = await env.run(msm_act.get_running_version)
+        # Verify version is in X.Y.Z format
+        parts = version.split(".")
+        assert len(parts) == 3
+        for part in parts:
+            assert part.isdigit()
```
MaikRe:
```suggestion
        # Verify version is in X.Y.Z format
        import re
        assert re.match(r'^\d+\.\d+\.\d+$', version)
```

src/tests/maastemporalworker/workflow/test_msm.py
```diff
@@ -373,14 +381,19 @@ async def test_send_heartbeat(self, mocker, msm_act, hb_param):
         assert kwargs["json"]["name"] == _MAAS_SITE_NAME
         assert kwargs["json"]["url"] == _MAAS_URL
         assert "machines_by_status" in kwargs["json"]
+        assert kwargs["json"]["version"] == hb_param.version
+        assert "known_config_options" not in kwargs["json"]
```
MaikRe, line 386: I might be missing something here but is there any test for asserting `known_config_options in kwargs["json"]` when `send_config_opts is True`?

MaikRe (commented):
Suggestion + Question inline

src/tests/maastemporalworker/workflow/test_msm.py
wyattrees: Technically no (that code path does get executed in a later test but nothing was asserted about it). I added an explicit activity test for this

MaikRe (approved):
+1

PR #181, opened 2026-04-22
fix(redfish): retrieval of correct Redfish node ID from System

r00ta:
Resolves LP:2141600


r00ta:
This patch has been tested successfully in TOR3 with `rockman`

troyanov (approved):

PR #182, opened 2026-04-22
fix: accept lists and strings for expected audience when decoding JWT

wyattrees:
Per RFC 7519, a JWT's `aud` claim can be a list of strings or a string, so enable checking for both when decoding an unverified JWT.

### DIFF
--- a/src/maascommon/utils/jwt.py
+++ b/src/maascommon/utils/jwt.py
@@ -30,7 +30,7 @@ class JWTAudienceError(JWTDecodeError):
 def decode_unverified_jwt(
     token: str,
     check_expiration: bool = True,
-    expected_audience: str | list[str] | None = None,
+    expected_audience: str | None = None,
 ) -> dict[str, Any]:
     """Decode a JWT without verifying its signature to extract claims."""
     try:
@@ -45,11 +45,13 @@ def decode_unverified_jwt(
         if expected_audience is not None:
             aud = claims.get("aud")
             throw = False
-            match expected_audience:
+            match aud:
                 case str():
                     throw = aud != expected_audience
                 case list():
-                    throw = aud not in expected_audience
+                    throw = expected_audience not in aud
+                case None:
+                    throw = True
             if throw:
                 raise JWTAudienceError(
                     f"invalid audience: expected {expected_audience}, got {aud}"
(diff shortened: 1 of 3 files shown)


alexsander-souza (dismissed):
+1

src/maascommon/utils/jwt.py
```diff
@@ -30,7 +30,7 @@ class JWTAudienceError(JWTDecodeError):
 def decode_unverified_jwt(
     token: str,
     check_expiration: bool = True,
-    expected_audience: str | None = None,
+    expected_audience: str | list[str] | None = None,
```
r00ta: `str | list[str]` is not a good type for an argument, I'd just use `list[str]`

r00ta (changes requested):

src/maascommon/utils/jwt.py
wyattrees: How do you suggest checking the `aud` claim when we expect it to be a `str` and not `list[str]`?

wyattrees: Nevermind, I refactored this to always take `expected_audience` as a `str` and check the type of the `aud` provided in the token.

r00ta: well we control what we pass as argument to this function. If you just keep the old typing `str | None` but you manage `list[str]` internally it is even worse!

Just keep `list[str]` and pass `[value]` when you want to use a single `str`?

wyattrees: So if a token has a single audience specified as `"aud": "some-audience"` as a claim, we just check `"some-audience" in expected_audience`? Or, do we not care if a token gives a single string as the `aud` and just fail? Or, do we wrap any `aud` present in the token as a single string in a list and compare it directly to `expected_audience`?

The intent behind having `expected_audience` being a `str` is that the token can specify several audiences that are valid, and we're checking here that the one we care about is accepted, no matter how many accepted audiences exist. If we want to match `expected_audience` to what is in the token exactly, then I think it being of type `str | list[str]` makes sense because that is the type of the `aud` claim per the RFC referenced above, unless we want to explicitly reject any token that has the `aud` claim as a string and not a list.

r00ta: `expected_audience` is something that we control. If it was coming directly from the JWT (and we can't control it) then it would make sense

r00ta: For example, if you have `expected_audience=x` but in the jwt token you get `aud=[x]`, would you match it?

r00ta: your latest version looks good, you keep `expected_audience` always as a single string and you manage the case where `aud` is a string or a list of strings

r00ta (approved):

PR #184, opened 2026-04-22
chore: remove sync_boot_source_selections_from_msm to prep for implementation of site_profiles

MaikRe:
Currently, when a Site enrolls with Site Manager, it will update its selections so that it will download every image offered by Site Manager. Image selections will be managed by Site Profiles moving forward, so this behavior should be removed.


wyattrees (approved):
+1

PR #186, opened 2026-04-23
chore(ci): tics job outside container

troyanov:
This change splits the TiCS workflow into two jobs and moves coverage generation out of the TiCS analyzer container/job.
This makes things easier, since we can run our tests in the dev container that is required by MAAS, and then do the TiCS analysis in a recommended way.


alemar99 (approved):
Nice job!

PR #188, opened 2026-04-23
fix: change expected bin path refs for core26

AloizioMacedo:
With core26, we changed paths of binaries from `/bin` to `/usr/bin`. We have several references to the `/bin` path in the code, and this PR changes them to appropriately reflect the expected ones now.

Note that it seems this does not affect snap builds due to symlinks. However, it affects the dev-snap when running `snap try` (as in there the symlinks do not seem to apply).


alexsander-souza (approved):
+1

PR #189, opened 2026-04-24
fix: calculate the sync percentage correctly for multiple regions

alemar99:
When multiple regions are present, the sum(bootresourcefile.size) was multiplying the file size * number of regions. To avoid this, we pre-group-by the sync sizes for each file.

Also show the status for selections which don't have a boot source cache associated anymore. This is to deal with the possibility in v2 to change the boot source URL, resulting in a synchronized selection but with no corresponding cache.


alexsander-souza (commented):
did you account for `bootresourcefilesync` being populated only after the download actually starts? we had issues in the past when a region was down or delayed, and the progress percentage was unstable

alemar99:
The progress percentage is still unstable and it's actually managed via the UI with optimistic updates. If a region is down or unstable I think it's correct to return an incomplete percentage as we already do

alemar99:
@alexsander-souza can I get a re-review?

alexsander-souza (approved):

PR #190, opened 2026-04-24
docs: add lxd to dependencies in maas versions

AloizioMacedo:
Updates things to synchronize with the contents of https://canonical.com/maas/docs/supported-maas-versions

### DIFF
--- a/docs/reference/supported-maas-versions.md
+++ b/docs/reference/supported-maas-versions.md
@@ -6,10 +6,10 @@ The table below lists dependencies for each MAAS release, including the base Ubu
 
 | Supported MAAS Versions | Dependencies  | Release Date | Type of Support |
 | - | - | - | - |
-| 3.7 | Ubuntu 24.04</br>PostgreSQL 16 |  December 2025  | Fully Supported |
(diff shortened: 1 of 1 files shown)


alexsander-souza (commented):
LXD is not a MAAS dependency. I think this would fit better in a compatibility section or something similar.

MaikRe (approved):
+1, I think for now this is fine until we have a compatibility docs page.

AloizioMacedo:
> LXD is not a MAAS dependency. I think this would fit better in a compatibility section or something similar.

> +1, I think for now this is fine until we have a compatibility docs page.

This is something that @jurekh pointed out as a discrepancy between the current docs and the new ones. Since the syncing of the docs makes it not clear as to what is the intended state., I will check back with him.

AloizioMacedo:
After talking with @jurekh , he mentioned that this was understood as a nice place of visibility for field engineers and others, even if not strictly correct in the naming of "dependency". We are also doing essentially the same thing by having postgres in that table.

We might revisit this later. For instance, there will be an upcoming endeavor to clean up the content of the docs, particularly around the offline docs being self-contained.

But for keeping with the expectations of that table let's merge as is.

PR #193, opened 2026-04-24
refactor: migrate UUID to ULID for better timestamp handling

alexsander-souza:
Replace UUID-based beaconing with ULID (Universally Unique Lexicographically
Sortable Identifiers) to fix the timestamp extraction bug and improve protocol
robustness. ULIDs provide built-in millisecond precision timestamps without the
libuuid timestamp offset bug (LP#2145734), simplifying timestamp extraction and
allowing tests to pass reliably.

Maintain backward compatibility during transition: version 1 (UUID) packets are
silently dropped on the receive path, allowing mixed-version clusters to operate
safely. Version 2 (ULID) is the new PROTOCOL_VERSION.

Resolves LP:2145734


required-packages/base
```diff
@@ -71,6 +71,7 @@ python3-tempita
 python3-temporalio
 python3-txtftp
 python3-tz
+python3-ulid
```
tanzin8r, line 74: Do we need to consider older series? if yes python3-ulid is not available in the Noble apt archives. This will break make install-dependencies and any CI that still targets Noble or Jammy.

src/provisioningserver/utils/tests/test_services.py
```diff
@@ -1149,11 +1149,23 @@ def test_skips_creating_listen_port_when_run_with_fake_reactor(self):
         result = protocol.stopProtocol()
         self.assertIsNone(result)
 
+    def test_beaconReceived_drops_v1_packet(self):
+        """beaconReceived() must silently return for version-1 (UUID-based) packets."""
+        protocol = BeaconingSocketProtocol(Clock(), port=0)
+        beacon_json = {
+            "version": 1,
+            "type": "solicitation",
+            "payload": {"uuid": str(ULID()), "type": 1},
```
tanzin8r: v1 packet would carry a real UUID1 string, not a ULID. The test should use a real UUID1 string (e.g. str(uuid.uuid1())) so it accurately reflects what a v1 peer would actually send.

src/provisioningserver/utils/services.py
```diff
@@ -386,6 +387,26 @@ def set_ipv6_multicast_source_ifindex(sock, ifindex):
...
+    :param interface_address: The IPv4 address of the interface to join on.
+    """
+    ipv4_join_sockopt_args = socket.inet_aton(
+        BEACON_IPV4_MULTICAST
+    ) + socket.inet_aton(interface_address)
+    try:
+        sock.setsockopt(
+            socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP, ipv4_join_sockopt_args
+        )
+    except OSError:
```
tanzin8r: The bare except OSError: pass swallows all join failures silently. maybe it would be nice to add a comment stating which errors are expected here (e.g. EADDRINUSE for already-joined), and add a debug-level log for unexpected codes so field issues are diagnosable.

src/provisioningserver/utils/services.py
```diff
@@ -825,16 +844,19 @@ def beaconReceived(self, beacon_json):
             from the external tcpdump-based process, or from the sockets layer
             (with less information about the received packet).
         """
+        # Silently drop v1 (UUID-based) packets; v2 nodes do not process them.
+        if beacon_json.get("version", 1) < PROTOCOL_VERSION:
```
tanzin8r, line 851: Dropping v1 packets silently means a v2 rack controller will produce zero topology hints when paired with any v1 rack controller still in the cluster during a rolling upgrade. Hence rolling upgrades must be all-at-once for beaconing. Maybe a debug log here would be great so operators can diagnose why topology hints are missing during transitions.

tanzin8r (changes requested):
moving towards ULID is a good direction maybe consider dual-read transition. So, write  always send v2 ULID beacons and read accepts both. we can remove v1 read path after rollout window. Left a couple of comments.

required-packages/base
alexsander-souza: no, this fix won't be back-ported

src/provisioningserver/utils/tests/test_services.py
alexsander-souza: fair, but we want to drop the `uuid` module, so we should not call it here. I'm going to replace with a random string, this field won't be parsed anyway

tanzin8r (approved):
+1

PR #197, opened 2026-04-27
fix(ci): do not embed raw event into shell script

troyanov:
If the event payload contains characters that interfere with shell parsing, the generated script can break.


alemar99 (approved):
+1

PR #198, opened 2026-04-27
feat: add security logging for tokens to rack controller v2/agent

MaikRe:
Implementation for MA257

### DIFF
--- a/src/maasapiserver/v3/api/internal/handlers/agent.py
+++ b/src/maasapiserver/v3/api/internal/handlers/agent.py
@@ -213,11 +213,14 @@ class AgentHandler(Handler):
             query=query_bootstraptoken
         )
         if bootstraptoken is None or utcnow() >= bootstraptoken.expires_at:
-            logger.info(
-                f"{AUTHN_TOKEN_REUSED}:bootstraptoken",
-                type=SECURITY,
-                token_hash=hash_token_for_logging(agent_enroll_request.secret),
-            )
+            if bootstraptoken:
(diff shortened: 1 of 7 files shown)


src/maasapiserver/v3/api/internal/handlers/agent.py
```diff
@@ -205,6 +213,11 @@ async def enroll_agent(
             query=query_bootstraptoken
         )
         if bootstraptoken is None or utcnow() >= bootstraptoken.expires_at:
```
tanzin8r, line 215: If you want to cover both cases, Maybe split into two logs with clearer names "no row matched this secret,” and other for “row matched but expired". since "reused” doesn’t describe the first case at all.

src/maascommon/logging/security.py
```diff
@@ -26,3 +42,9 @@
 CREATED = "created"
 UPDATED = "updated"
 DELETED = "deleted"
+
+# Tokens
+AUTHN_TOKEN_CREATED = "AUTHN_token_created"
+AUTHN_TOKEN_DELETED = "AUTHN_token_deleted"
+AUTHN_TOKEN_REVOKED = "AUTHN_token_revoked"
```
tanzin8r, line 63: Where is this getting used?

MaikRe: Good point, this currently isn't used but it will be in upcoming PRs for more token logging.

troyanov (approved):

PR #204, opened 2026-04-28
ci: make python code changes trigger docs CI

AloizioMacedo:
This is necessary due to the automatic API and CLI generation.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -64,6 +64,11 @@ jobs:
               - 'src/**/go.sum'
             docs:
               - 'docs/**'
+              - 'src/**/*.py'
+              - 'utilities/generate_builders.py'
+              - 'utilities/run-py-tests-ci'
+              - 'pyproject.toml'
+              - 'tox.ini'
 
   lint:
     runs-on: ubuntu-24.04


.github/workflows/ci.yaml
```diff
@@ -64,6 +64,11 @@ jobs:
               - 'src/**/go.sum'
             docs:
               - 'docs/**'
+              - 'src/**/*.py'
+              - 'utilities/generate_builders.py'
```
alexsander-souza, line 68: changes in the builder are going to be reflected in updated generators in `src/**/*.py`, so this line is redundant.

.github/workflows/ci.yaml
```diff
@@ -64,6 +64,11 @@ jobs:
...
+              - 'utilities/run-py-tests-ci'
```
alexsander-souza, line 69: why this affects the docs?

.github/workflows/ci.yaml
```diff
@@ -64,6 +64,11 @@ jobs:
...
+              - 'pyproject.toml'
+              - 'tox.ini'
```
alexsander-souza, line 72: add `required-packages/base` (and maybe `snap/snapcraft.yaml`) so you run doc tests when we (potentially) modify the project dependencies

.github/workflows/ci.yaml
AloizioMacedo: The major thing is `src/**/*.py`. The rest is there for consistency with the `python` filters defined just above (in the ci.yaml).

It affects the docs because we use the source code to generate the API reference, and the CLI reference. So, for example, one thing that might happen is the following scenario: 
- Someone removes something from the API
- Part of the (static) docs refer to that part of the API by path reference
- The docs would raise a warning about the inconsistency, which would be a hard error for the docs build process in RTD. But since nothing changed in the static docs, without this PR this would be invisible in the CI.

Another example is #193. If the docs test ran with the change in the code, the docs checks would raise a missing dependency. Once we use MAAS as a local package for the docs this particular example will be further mitigated, but the point of this PR still stands due to the example mentioned above that doesn't involve dependencies and is a documentation reference issue.

.github/workflows/ci.yaml
AloizioMacedo: I see. This should also apply to the `python` filter then, right?

I won't change it in this PR to keep the scope, but just confirming my understanding.

AloizioMacedo:
Continuing in #205.

PR #206, opened 2026-04-28
feat: add Temporal activity to set global configuration from Site Manager profile

wyattrees:


src/maasservicelayer/db/repositories/database_configurations.py
```diff
@@ -60,3 +62,29 @@ async def create_or_update(
...
+        """
+        Set many configuration items at once. This method cannot be called when
+        the database connection is a psycopg2 connection handled by Django.
+        Args:
+            configuration: The configuration to set in the database. Keys of
+            this dict are the names of the configuration items.
+        """
+        connection = self.context.get_connection()
+        if isinstance(connection, Connection):
+            raise RuntimeError(
```
alemar99: Why it cannot be executed by a psycopg2 connection

src/maasservicelayer/services/configurations.py
```diff
@@ -181,6 +181,26 @@ async def set(
...
+        try:
+            if any(
+                [
+                    ConfigFactory.parse(name, value).stored_as_secret
+                    for name, value in configuration.items()
+                ]
+            ):
+                raise RuntimeError(
+                    "clear_and_set_many does not support setting secrets."
+                )
```
alemar99: It would probably be better to raise a ValidationException or something like that, specifying the field that cannot be passed

src/maasservicelayer/services/configurations.py
```diff
@@ -181,6 +181,26 @@ async def set(
...
+        except ValueError as err:
+            raise RuntimeError(
+                "Some configuration options are unknown or invalid."
+            ) from err
```
alemar99: Same as above

src/maastemporalworker/workflow/msm.py
```diff
@@ -313,6 +315,13 @@ async def set_selections(self, input: MSMSetSelectionsParam) -> None:
             )
             await services.boot_source_selections.create_many(builders)
 
+    @activity_defn_with_context(name=MSM_SET_GLOBAL_CONFIG_ACTIVITY_NAME)
+    async def set_global_config(self, input: MSMSetGlobalConfigParam) -> None:
+        async with self.start_transaction() as services:
+            await services.configurations.clear_and_set_many(
+                input.configuration
+            )
```
alemar99: Here you don't take into account exceptions and the activity will fail forever

alemar99 (changes requested):
See inline

src/maasservicelayer/db/repositories/database_configurations.py
wyattrees: Actually I think it's fine, I misunderstood how these connection types are used in MAAS. I changed this to allow it.

src/maastemporalworker/workflow/msm.py
wyattrees: These exceptions should never be hit here. We do validation on the values on the Site Manager side, so no invalid values or nonexistent settings should be passed, nor should any secrets be passed. However, if we still want to handle them here just in case, what do you recommend? Catching and raising an `ApplicationError`?

alemar99: Yeah I imagined that validation happens on another layer. Yes I'd go with `ApplicationError` with `non_retryable=True`

src/maasservicelayer/db/repositories/database_configurations.py
```diff
@@ -60,3 +62,27 @@ async def create_or_update(
...
+        connection = self.context.get_connection()
+        stmt = pg_insert(ConfigTable).returning(ConfigTable)
+        data = [
+            {"name": key, "value": value}
+            for key, value in configuration.items()
+        ]
+        if isinstance(connection, Connection):
+            result = connection.execute(stmt, data)
+        else:
+            result = await connection.execute(stmt, data)
```
alemar99: You should use `self.execute_stmt` which already take the connection type into account.

Since execute only accepts a statement, add the data to the statement before:
```python
data = [
    {"name": key, "value": value}
    for key, value in configuration.items()
]
stmt = pg_insert(ConfigTable).values(data).returning(ConfigTable)
```
(not tested)

alemar99 (changes requested):

src/tests/maasservicelayer/conftest.py
```diff
@@ -10,6 +11,7 @@
 __all__ = [
     "db",
     "db_connection",
+    "db_connection_sync",
```
alemar99: remove

src/tests/maasapiserver/fixtures/db.py
```diff
@@ -133,6 +134,14 @@ def random_string(self, length: int = 10) -> str:
         return "".join(random.choices(string.ascii_letters, k=length))
 
 
+@pytest.fixture
+async def db_connection_sync(
```
alemar99: remove

alemar99 (dismissed):
LGTM

tanzin8r (approved):
+1

PR #208, opened 2026-04-28
fix: coerce memory into int for machine listing

AloizioMacedo:
Pydantic v2 does not seem to coerce a decimal into an int implicitly, and we relied on this before.

This change preserves the way the machines listing was displayed before we updated our dependencies.


AloizioMacedo:
For reference, 

- This is the current state if a machine has RAM indicated as 31.75 (my case):

  image

- This is after the change:

  image

- This is in 3.7 (ignore the extraneous test machine):
  
  image

alemar99 (approved):
+1

PR #209, opened 2026-04-29
docs: refactor get maas up and running

tobiasdemendonca:
This PR refactors the `get-maas-up-and-running` how-to to facilitate a "simple MAAS installation".

The changes boil down to splitting it up into these documents: 

- Explain an overview of installation at the new `Get started` index.md. This includes pointing to a the simple MAAS installation, how to configure postgresql for remote connections, and finally to `maas-terraform-modules` for HA Charmed MAAS
- Install MAAS
   - Remove reference to `maas-test-db` - in theory this is simple enough to setup without requiring it, and removes confusion about what "production" vs "POC" is for people like home-labbers
  - Document using the Postgresql snap, and move the deb install of MAAS to a less obvious location to discourage users from using it.
  - Changes to the postgresql connection to use localhost, which saves having to add `md5` authentication to `pg_hba.conf` for the maas user/db. This by default uses the more secure `scram-sha-256` password encryption)
  - streamline admin user creation
- Configure MAAS
  - at the moment this is DNS and DHCP, but could be extended
  - DHCP setup was pretty confusing before. I've attempted to be more verbose and completely revised it to reflect the thought process a user would actually have to go through. 
- Upgrade MAAS - moved out of the original `get-maas-up-and-running` to the root how-to level as a separate document. This needs to be reworked as it lacks sufficient detail. 
- Configure postgresql for remote connections - added as a bonus for those who need it. This wasn't completely covered by the original doc as it missed adding a listen address, so this more of an addition than a fix. 

To view the docs locally, run `make serve` in `docs/`


tobiasdemendonca:
Current question marks include: 

- If disabling conflicting ntp servers is still required?
- install requirements previously included the os as jammy. Is this accurate or do we actually just need snapd on any ubuntu system to install the snap, as we have the relevant core-xx base? 
- The old troubleshooting section is quite vague, currently still in `getting-started/index.md`. I think the first point can be removed, but is the bmc troubleshooting step worth keeping?

troyanov:
> If disabling conflicting ntp servers is still required?

I never disable them and works fine on my machine. I think it would be great to check with FE folks what is their most recent experience.

docs/reference/configuration-guides/installation-requirements.md
```diff
@@ -50,7 +50,7 @@ Too few database connections tend do produce errors:
 
 ## MAAS, NTP, and chrony
 
-Conflicts can arise between Ubuntu's default `systemd-timesyncd` and MAAS `chrony`. Consult the MAAS [installation guide](/how-to-guides/get-maas-up-and-running.md) for solutions.
+Conflicts can arise between Ubuntu's default `systemd-timesyncd` and MAAS `chrony`. Consult the MAAS [installation guide](/how-to-guides/get-started/install-maas.md) for solutions.
```
troyanov, line 53: This needs to be validated with FE folks.

docs/reference/release-notes.md
```diff
@@ -1043,11 +1043,11 @@ The MAAS 3.4 release expands Packer support to include SUSE Linux Enterprise Ser
 
 MAAS will run on [just about any modern hardware configuration](/reference/configuration-guides/installation-requirements).
 
-- [How to do a fresh install of MAAS 3.4](/how-to-guides/get-maas-up-and-running): Use the tabs to select snaps or packages.
+- [How to do a fresh install of MAAS 3.4](/how-to-guides/get-started/index.md): Use the tabs to select snaps or packages.
```
troyanov, line 1046: > Use the tabs to select snaps or packages.

Is it still accurate? I don't think we have tabs anymore?

docs/how-to-guides/upgrading-maas.md
```diff
@@ -0,0 +1,42 @@
...
+- MAAS 3.3: PostgreSQL 12 deprecated. Upgrade to 14 before proceeding.
+- MAAS 2.8 or earlier: Full backup required. Fresh install recommended if upgrade fails.
+
+## Verification
+
+After upgrade:
+
+```bash
+lsb_release -a   # Verify Ubuntu release
+maas --version   # Verify MAAS version
```
troyanov: I think those two lines are redundant, as we only need to validate that MAAS works.

docs/how-to-guides/upgrading-maas.md
```diff
@@ -0,0 +1,42 @@
+# Upgrading MAAS
+
+This guide outlines the steps to upgrade MAAS to the latest version
+
+## General steps
+
+1. Backup your system and database.
+2. Verify Ubuntu release (`lsb_release -a`). Upgrade to 22.04 Jammy or 24.04 Noble as required.
```
troyanov: Maybe something more generic here?


```suggestion
2. Verify Ubuntu release (`lsb_release -a`) and upgrade to the newer one if required.
```

docs/how-to-guides/upgrading-maas.md
```diff
@@ -0,0 +1,42 @@
...
+3. Verify PostgreSQL version (14 required, 16 recommended).
```
troyanov:
```suggestion
3. Verify PostgreSQL version compatibility and upgrade it if required.
```

docs/how-to-guides/get-started/install-maas.md
```diff
@@ -0,0 +1,115 @@
+# Install MAAS
+
+There are multiple approaches to installing MAAS, depending on your needs, use case, and operating system. This document outlines how to install MAAS on a single machine in region+rack mode. As MAAS is packaged without a database, both the MAAS snap and the PostgreSQL snap are required.
```
troyanov: I would remove the part about PostgreSQL 

```suggestion
There are multiple approaches to installing MAAS, depending on your needs, use case, and operating system. This document outlines how to install MAAS on a single machine in region+rack mode.
```

docs/how-to-guides/get-started/install-maas.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Requirements
```
troyanov: We can list PostgreSQL as a requirement as well and mention that it can be installed via snap. What do you think?

docs/how-to-guides/get-started/install-maas.md
```diff
@@ -0,0 +1,115 @@
...
+
+To install older versions of MAAS, see [Supported MAAS versions](/reference/supported-maas-versions.md) for compatible PostgreSQL channels and operating system versions. For this guide, we will use the latest version of MAAS.
+```
+
+## Install the MAAS snap
+
+Install the MAAS snap with:
+
+```bash
+sudo snap install maas --channel=3.7/stable
```
troyanov: I am curious if we need to introduce `latest/stable` channel so the doc is always _up to date_
What do you think?

docs/how-to-guides/get-started/install-maas.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+## Install PostgreSQL
```
troyanov, line 30: I think this step is optional, someone might already have a PostgreSQL database up and running somewhere on the network.

docs/how-to-guides/get-started/install-maas.md
```diff
@@ -0,0 +1,115 @@
...
+Install the PostgreSQL snap:
+
+```bash
+sudo snap install postgresql --channel=16/stable
+```
+
+Configure credentials for the PostgreSQL database. Pick a strong password and store it securely:
+
+```bash
+export DBPASS=<strong-password>
```
troyanov: That already makes password insecure as it was exposed via ENV, I would rather propose a prompt approach here.

docs/how-to-guides/get-started/install-maas.md
```diff
@@ -0,0 +1,115 @@
...
+export DBUSER=maas
+export DBNAME=maas
+```
+
+Create a MAAS database role and database:
+
+```bash
+sudo postgresql.psql -U postgres -c "CREATE USER \"$DBUSER\" WITH ENCRYPTED PASSWORD '$DBPASS'"
+sudo postgresql.createdb -U postgres -O "$DBUSER" "$DBNAME"
```
troyanov:
```suggestion
postgresql.createuser -U postgres -h /tmp -P maas
postgresql.createdb -U postgres -h /tmp -O maas maas
```

docs/how-to-guides/get-started/install-maas.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+## Initialize MAAS
+
+Initialize MAAS in region+rack mode with the PostgreSQL database credentials:
```
troyanov:
```suggestion
Initialize MAAS in region+rack mode:
```

docs/how-to-guides/get-started/install-maas.md
```diff
@@ -0,0 +1,115 @@
...
+
+```bash
+sudo maas init region+rack --database-uri "postgres://$DBUSER:$DBPASS@localhost/$DBNAME"
```
troyanov:
```suggestion
sudo maas init region+rack --database-uri "postgres://<user>:<password>@<hostname>/<dbname>"
```

docs/how-to-guides/get-started/install-maas.md
```diff
@@ -0,0 +1,115 @@
...
+```bash
+sudo maas createadmin \
+  --username admin \
+  --email admin@localhost \
+  --ssh-import gh:<github-username>  # optional
+```
+
+Specify the password for your new administrator user when prompted. Store it securely. 
+
+<!--- do we need them to store this securely or can it be obtained later? --->
```
troyanov: We don't store password in clean text anywhere. Password can be reset if you have privileged access to `maas` CLI.


```
root@maas:~# maas changepassword --help
usage: maas changepassword [-h] [--database {default}] [username]

Change a MAAS user's password.

positional arguments:
  username              Username to change password for; by default, it's the current username.

options:
  -h, --help            show this help message and exit
  --database {default}  Specifies the database to use. Default is "default".
```

docs/how-to-guides/get-started/install-maas.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Verify your installation
+
+Verify the installation by creating a profile and logging in:
+
+```bash
+maas login admin "http://localhost:5240/MAAS/api/2.0/" $(sudo maas apikey --username=admin)
```
troyanov:
```suggestion
maas login admin "http://localhost:5240/MAAS" $(sudo maas apikey --username=admin)
```

I think that should be enough, no need for `/api/2.0/`

docs/how-to-guides/get-started/install-maas.md
```diff
@@ -0,0 +1,115 @@
...
+For help with the available commands, try:
+
+  maas admin --help
+
+```
+
+```{admonition} Profiles
+:class: note
+
+You have just created a profile in the MAAS CLI called `admin`, logged in using the credentials created in the previous step for your admin user. The rest of these docs will refer to this profile as `$PROFILE`.
```
troyanov:
```suggestion
You have just created a profile in the MAAS CLI called `admin`, logged in using the API key created in the previous step for your admin user. The rest of these docs will refer to this profile as `$PROFILE`.
```

docs/how-to-guides/get-started/install-maas.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+## Access the MAAS UI
+
+Open a browser and navigate to `http://<maas-ip>:5240/MAAS`, replacing `<maas-ip>` with the IP address of your MAAS machine. If you are on the same machine as the machine running MAAS, use `http://localhost:5240/MAAS`.
```
troyanov:
```suggestion
Open a browser and navigate to `http://<maas-ip>:5240/MAAS`, replacing `<maas-ip>` with the IP address of your MAAS region controller. If you are on the same machine as the machine running MAAS, use `http://localhost:5240/MAAS`.
```

troyanov (commented):
Nice work! 

I do have some suggestions/questions for the Installation part and I didn't have time to review Configure MAAS yet.

docs/reference/release-notes.md
tobiasdemendonca: This is just me fixing broken links. As this document is going to be on "latest", and not backported (because of 3.7 specific references), I don't think I should go back and correct all older MAAS version references in this PR

docs/how-to-guides/upgrading-maas.md
tobiasdemendonca: I have just split out Bills older work here. I can remove these though, but I think this document needs a more thorough revision in a separate PR

docs/how-to-guides/get-started/install-maas.md
tobiasdemendonca: This was on my list of points to discuss earlier actually: I think `latest/stable` would be great to have imo. But ultimately, we will always have to change the relevant documentation because of postgres version compatibility. 

We _could_ inject these in as variables in future, getting the MAAS version from `pyproject.toml`, but we'd need to add a more explicit variable for `PSQL_RECOMMENDED_VERSION`. We only have `PSQL_MIN_VERSION` at the moment in `plugin.py`, with 16 [hard-coded](https://github.com/canonical/maas/blob/8f4278e4e001daadf7e01446e20545279c08c6ad/src/maasserver/deprecations.py#L82) elsewhere

docs/reference/configuration-guides/installation-requirements.md
tobiasdemendonca: I believe this is relevant as this is in fce, here is the relevant bug: https://bugs.launchpad.net/maas/+bug/2003812

I think this is relevant only for ubuntu < 25.10 as after that it was replaced by chrony which we use in MAAS
https://ubuntu.com/server/docs/how-to/networking/timedatectl-and-timesyncd/

docs/how-to-guides/get-started/install-maas.md
tobiasdemendonca: I think the later note to skip the section if already installed covers it, otherwise we'll have an optional requirement

docs/how-to-guides/get-started/install-maas.md
tobiasdemendonca: Fixed, but this is much less "copy and paste" friendly. Other how-to instructions e.g. [aws](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html) are comfortable doing this though it's worth noting

docs/how-to-guides/get-started/install-maas.md
tobiasdemendonca: I have set user and dbname hardcoded to maas instead to align with your above suggestion, this way the user only needs to input one thing

docs/reference/configuration-guides/installation-requirements.md
tobiasdemendonca: Just FYI i've added a not in `install-maas.md` about this

troyanov: Ack. Thanks for double checking 👍

docs/how-to-guides/get-started/install-maas.md
troyanov: Good point about copy-paste friendly. Oki, let's not bother too much and address this to TA

troyanov (dismissed):
As we discussed during the call, this is a solid documentation improvement. I know neither of us are TAs, but I appreciate your sense for what makes things clearer and more useful.

docs/how-to-guides/get-started/index.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Troubleshooting notes
+
+- NTP conflicts:
+
+  ```bash
+  sudo systemctl disable --now systemd-timesyncd
+  ```
+
+- BMC migration (3.3+): Ensure unique BMC IP/username/password combinations.
```
skyecas: Q, not nit; did you ever find a document that goes into this in more detail?

skyecas (dismissed):
nice!

docs/how-to-guides/get-started/index.md
tobiasdemendonca: Thanks for this, I found the relevant bug https://bugs.launchpad.net/maas/+bug/2025026 and because of the context of this, have moved this to the upgrading doc. I removed the other troubleshooting step as this is now a prerequisite, so this is cleaner

troyanov (approved):

PR #214, opened 2026-04-29
chore(ci): better regex to find Launchpad bug ID

troyanov:


alemar99:
Looks good, but the conventional commits spec says that the footer must be separated either with `:<space>` or `<space>#`. Maybe we should follow the spec, and also I like more the idea of having only one keyword possible, wdyt?

troyanov:
> I like more the idea of having only one keyword possible

I am also fine with just a single word, the reason why I extended the list is mostly to match [GitHub behaviour](
https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests)

You are talking about this one, right?
> One or more footers MAY be provided one blank line after the body. Each footer MUST consist of a word token, followed by either a :<space> or <space># separator, followed by a string value (this is inspired by the git trailer convention).

So I didn't include `#` on purpose, because otherwise `#[0-9]+` will be identified as a GitHub issue or Pull Request, and [sometimes](https://github.com/canonical/maas/commit/a1f6ad325f0eeb85813b47dc3129e4346a6e8cc2) it is not a footer at all.

FTR here is a [playground](https://regex101.com/?regex=%5Cb%28%3F%3Aclose%7Ccloses%7Cclosed%7Cfix%7Cfixes%7Cfixed%7Cresolve%7Cresolves%7Cresolved%29%5Cb.*LP%3A%5Cs*%5CK%5B0-9%5D%2B&testString=fix%28redfish%29%3A+retrieval+of+correct+Redfish+node+ID+from+System+%28%23181%29%0AResolves+LP%3A2141600&flags=gmi&flavor=pcre&delimiter=%2F) for the provided regex.

alemar99 (approved):

PR #215, opened 2026-04-29
docs: transition our docs url

AloizioMacedo:
Most of the process here will just follow https://canonical-rtd-proxy.readthedocs-hosted.com/how-to/migrate/

### DIFF
--- a/docs/_static/js/overwrite_links.js
+++ b/docs/_static/js/overwrite_links.js
@@ -1,0 +1,23 @@
+// Replace oldDomain with newDomain
+const oldDomain = 'canonical-maas.readthedocs-hosted.com';
+const newDomain = 'canonical.com/maas/docs';
+
+// Use a MutationObserver to wait for the RTD flyout element to appear in the DOM
+const observer = new MutationObserver(function (mutations, obs) {
+  const rtdFlyout = document.querySelector('readthedocs-flyout');
+  if (!rtdFlyout) return;
+
+  obs.disconnect();
+
+  rtdFlyout.addEventListener('click', function () {
+    const shadowRoot = rtdFlyout.shadowRoot;
+    if (!shadowRoot) return;
+
+    const anchors = shadowRoot.querySelectorAll('a');
+    anchors.forEach(anchor => {
+      anchor.href = anchor.href.replace(new RegExp(oldDomain, 'g'), newDomain);
+    });
+  });
+});
+
+observer.observe(document.body, { childList: true, subtree: true });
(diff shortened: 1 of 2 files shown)


AloizioMacedo:
Now that the [docs are in staging](https://staging.canonical.com/maas/docs/latest/), this PR can be merged and materially tested. It should fix things like the redirect in the readthedocs version flyout.

For reference, this is [part of the migration steps](https://canonical-rtd-proxy.readthedocs-hosted.com/how-to/migrate/#configure-a-branch-of-your-documentation).

docs/conf.py
```diff
@@ -220,15 +220,15 @@ def dummy_setup(app):
 # TODO: If your documentation is hosted on https://docs.ubuntu.com/,
 #       uncomment and update as needed.
 
-# slug = ''
+slug = "maas/docs"
```
tmerten, line 223: I don't know their definition of a slug but from what I know it should be stuff between to slashes `/`: `/this-is-my-slug/this-is-another/`. But maybe RTD's definition is different. This looks like a prefix to me.

docs/_static/js/overwrite_links.js
```diff
@@ -0,0 +1,23 @@
+// Replace oldDomain with newDomain
```
tmerten, line 1: Possibly the right call but does this need to be applied dynamically? Can't we search replace on static content before hosting? Or is this a safety net?

tmerten (changes requested):

alexsander-souza (approved):

docs/_static/js/overwrite_links.js
AloizioMacedo: Good question!

So, the issue is that the rtd flyout is not something that is ever in our static files. You can see this materially by running the docs locally and comparing with RTD:

- Locally:
image

- RTD:
image

So it needs to be applied dynamically : | (even though this is not really explained in the docs of the migration)

docs/conf.py
AloizioMacedo: This is based on the guidance of https://canonical-rtd-proxy.readthedocs-hosted.com/how-to/migrate/, specifically this part:

> slug = '\<product\>/docs' # Or '\<ecosystem\>/\<product\>/docs'
> 
> html_baseurl = f"https://\<\<NewURL\>\>/"
> 
> ogp_site_url = f"https://\<\<NewURL\>\>/"
>
> html_static_path = ["_static"]
> 
> html_js_files = [
>    "js/overwrite_links.js",
>]

tmerten (approved):
thanks for the explanations!

PR #216, opened 2026-04-29
feat: add token logging for msm enrollment

MaikRe:
Implementation for MA257

### DIFF
--- a/src/maasservicelayer/services/msm.py
+++ b/src/maasservicelayer/services/msm.py
@@ -9,6 +9,11 @@ import structlog
 from temporalio.common import WorkflowIDReusePolicy
 
 from maascommon.enums.msm import MSMStatusEnum
+from maascommon.logging.security import (
+    AUTHN_TOKEN_REVOKED,
+    hash_token_for_logging,
+    SECURITY,
+)
 from maascommon.utils.jwt import decode_unverified_jwt, JWTDecodeError
 from maascommon.workflows.msm import (
     MSM_ENROL_SITE_WORKFLOW_NAME,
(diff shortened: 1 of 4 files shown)


src/maastemporalworker/workflow/msm.py
```diff
@@ -440,11 +466,30 @@ async def refresh_token(
...
+                        msm_url=input.sm_url,
+                    )
                     return (
-                        data["access_token"],
+                        new_token,
                         data["rotation_interval_minutes"],
                     )
                 case 401 | 404:
+                    logger.info(
+                        f"{AUTHN_TOKEN_REUSED}:MSM:accesstoken",
```
wyattrees, line 488: wouldn't `AUTHN_TOKEN_REVOKED` make more sense here?

src/maasservicelayer/services/msm.py
```diff
@@ -236,6 +241,14 @@ async def withdraw(self) -> None:
         )
         if not msm_creds:
             return None
+
+        logger.info(
+            f"{AUTHN_TOKEN_REVOKED}:MSM:accesstoken",
```
wyattrees, line 246: I'm not sure if `REVOKED` or `DELETED` makes more sense here. Technically the token is still valid, we're just choosing not to use it anymore.

wyattrees (commented):
A few questions in-line. The notion of `REVOKED`/`DELETED`/etc is a bit different here, since MAAS is a client and not the provider of the JWT, so I'm not sure if the constants exactly apply to these situations or if we need new constants

MaikRe:
Closing this in favor of logging these in MSM in the future since MAAS is the client in the enrollment case and is not responsible for the lifetime of these tokens.

PR #218, opened 2026-04-30
feat(ci): Auto-PR to MAAS UI for API contract changes MAASENG-6451

abuyukyi101198:
## Done

- Added `api-client-sync.yaml` workflow

## What it does

The workflow
- runs on pushes to `master` or `api-client-sync-debug-*` in `canonical/maas`
- checks out both `maas` and `maas-ui`
- builds `master` branch MAAS, and sets up an LXD MAAS instance
- downloads the OpenAPI spec, and exposes it as a GH Action artifact
- runs API client generation on the downloaded spec using MAAS UI's `main` configuration
- detects if there is an existing API client update branch in `canonical/maas-ui`, if not, creates one
- pushes the updated client files to the branch
- opens a PR to `main` in `canonical/maas-ui`

Here is an [example PR](https://github.com/canonical/maas-ui/pull/6004) generated by this workflow.

### DIFF
--- a/.github/workflows/api-client-sync.yaml
+++ b/.github/workflows/api-client-sync.yaml
@@ -116,3 +116,52 @@ jobs:
       - name: Generate API client
         working-directory: fe-repo
         run: yarn generate-api-client
+
+      - name: Restore openapi-ts config
+        working-directory: fe-repo
+        run: git checkout openapi-ts.config.ts
+
+      - name: Create PR branch and push
+        id: push-branch
+        working-directory: fe-repo
+        run: |
+          git config user.name "${{ vars.LANDER_USERNAME }}"
+          git config user.email "${{ vars.LANDER_EMAIL }}"
+
+          BRANCH="chore/update-api-client-${{ github.sha }}"
+          git checkout -b "$BRANCH"
+
+          git add src/app/apiclient/
+          
+          if git diff --cached --quiet; then
+            echo "No changes detected in the API client files, skipping PR."
+            echo "has_changes=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          git commit -m "chore: regenerate API client
+
+          Triggered by: ${{ github.repository }}@${{ github.sha }}"
+          git push origin "$BRANCH"
+          echo "branch=$BRANCH" >> "$GITHUB_OUTPUT"
+          echo "has_changes=true" >> "$GITHUB_OUTPUT"
+
+      - name: Open PR on MAAS UI repo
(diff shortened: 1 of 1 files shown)


.github/workflows/api-client-sync.yaml
```diff
@@ -0,0 +1,181 @@
...
+      - name: Setup LXD
+        uses: canonical/setup-lxd@v1
+        with:
+          channel: 5.21/stable
+
+      - name: Build MAAS snap
+        run: |
+          sudo snap install snapcraft --classic
+          sudo snap refresh snapd
+          sudo snap install --edge core26
```
ndv99: This should use candidate instead of edge; it'll be more stable and still work the same for our use case
```suggestion
          sudo snap install --candidate core26
```

.github/workflows/api-client-sync.yaml
```diff
@@ -0,0 +1,181 @@
...
+          make snap
+          make snap-clean
+
+      - name: Launch and configure LXD container
+        id: setup-lxd
+        run: |
+          lxc launch ubuntu-daily:resolute maas-api-client-gen
```
ndv99: Resolute images are now available in LXD

```suggestion
          lxc launch ubuntu:resolute maas-api-client-gen
```

.github/workflows/api-client-sync.yaml
```diff
@@ -0,0 +1,181 @@
...
+          lxc exec maas-api-client-gen -- cloud-init status --wait
+          lxc exec maas-api-client-gen -- snap install snapd --beta
+          lxc exec maas-api-client-gen -- snap wait system seed.loaded
+          lxc exec maas-api-client-gen -- snap install --edge core26
```
ndv99:
```suggestion
          lxc exec maas-api-client-gen -- snap install --candidate core26
```

.github/workflows/api-client-sync.yaml
```diff
@@ -0,0 +1,181 @@
...
+        run: yarn install --frozen-lockfile
+
+      - name: Update openapi-ts config to use local spec
+        working-directory: fe-repo
+        run: |
+          python3 -c "
+          import re, pathlib
+          p = pathlib.Path('openapi-ts.config.ts')
+          p.write_text(re.sub(r'input:.*', \"input: './openapi-spec.json',\", p.read_text()))
+          "
```
ndv99, line 110: I think we should adjust this directly in MAAS UI rather than patching it on the fly here, but this can be done separately

ndv99 (commented):
Looking mostly good, some corrections below

ndv99 (approved):
LGTM

PR #219, opened 2026-04-30
chore(ci): enforce conventional commits as per MA201 spec

alemar99:


alemar99:
We might want to revise the scopes now or later.
I've made the bug reference mandatory for `fix` commits even if the spec says "It should contain a bug reference"

alemar99:
Action works correctly works on `edited` events.
I've changed the title to `feat(ci): ...` and this is the associated action run: https://github.com/canonical/maas/actions/runs/25159676723/job/73750519252?pr=219

alemar99:
Here is another run for `fix: ...` commit title: https://github.com/canonical/maas/actions/runs/25159743810/job/73750751104?pr=219

AloizioMacedo:
> We might want to revise the scopes now or later. I've made the bug reference mandatory for `fix` commits even if the spec says "It should contain a bug reference"

I think adhering to the spec is a better call here. There are fixes that are not tied to a bug report (e.g. #208), and making it mandatory would only add friction and unnecessary processes.

We could instead, as a middle ground, make an automated reply from GitHub in case it is missing. Something like

```If your PR addresses a bug in LP, please make sure to add the correct footer: Resolves LP:X.```

This also keeps with the spirit of the spec as written. (I think the above is feasible.)

alemar99:
Closing as feature was already added.

PR #222, opened 2026-04-30
chore: update user stats endpoints to reflect new API guidelines

ndv99:
## Done
- /users_with_summary changed to /users/statistics
- /me_with_summary changed to /me/with/statistics
- removed duplicate information in endpoints (except for ID)

### DIFF
--- a/src/maasapiserver/v3/api/public/models/requests/users.py
+++ b/src/maasapiserver/v3/api/public/models/requests/users.py
@@ -20,11 +20,14 @@ class UsersFiltersParams(BaseModel):
             description="Filter by User ID",
         )
     )
+
     username_or_email: str | None = Field(
         Query(default=None, title="Filter by username or email")
     )
 
     def to_clause(self) -> Clause | None:
+        if self.ids:
+            return UserClauseFactory.with_ids(self.ids)
         if self.username_or_email:
             return UserClauseFactory.with_username_or_email_like(
                 self.username_or_email
--- a/src/maasapiserver/v3/api/public/models/requests/users.py
+++ b/src/maasapiserver/v3/api/public/models/requests/users.py
@@ -13,6 +13,13 @@ from maasservicelayer.models.base import UNSET
 
 
 class UsersFiltersParams(BaseModel):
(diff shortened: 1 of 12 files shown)


ndv99:
struggling to get the dev setup working to run checks locally. will investigate more tomorrow

nehjoshi (changes requested):
For the `/users:statistics` endpoint, we want the request URL to accept a list of user ids or emails (ids would probably be better, since we've used that convention for most statistics endpoints, and because a long list of emails will blow up the url size). 

Right now, it only accepts `username_or_email` which will just filter out statistics for a single user. For passing in multiple ids as a filter, you can take a look at `src/maasapiserver/v3/api/public/models/requests/usergroups.py` (lines 15-21)

src/maasapiserver/v3/api/public/models/requests/users.py
```diff
@@ -13,6 +13,13 @@
 
 
 class UsersFiltersParams(BaseModel):
+    ids: list[int] | None = Field(
+        Query(
+            default=None,
+            alias="id",
+            description="Filter by User ID",
+        )
+    )
```
nehjoshi, line 22: You'll also have to parse the received IDs in the `to_clause` and `to_href_format` functions in the same file so that they are sent to the service layer and then the DB query. For this, you might need to create a `UserClauseFactory.with_ids()` function in the user repository.

nehjoshi (changes requested):

src/maasapiserver/v3/api/public/models/requests/users.py
nehjoshi: You can refer to the entire `usergroups.py` suite (handler, service and repository) and it should have a similar implementation

ndv99: Thanks! That really helped. I think this commit I've added should do it, doesn't seem like the service needed any changes, just the repository and handler.

nehjoshi (approved):
LGTM, let's gooo!

PR #223, opened 2026-04-30
docs: add contextual logging docs

MaikRe:


tobiasdemendonca (commented):
I have a general comment on this as I've been looking into the getting started docs. 

I don't think this is a how-to doc, it's more of an informational doc ("how to contextual logging" doesn't seem to fit with the rest of the how-to titles, and also the content). Even though you've nested it under `use-logging`, imo it should still be a `how-to` type doc (e.g. [lxd docs](https://documentation.ubuntu.com/lxd/default/howto/).

 Isn't this more like a reference or explanation?

MaikRe:
> I have a general comment on this as I've been looking into the getting started docs.
> 
> I don't think this is a how-to doc, it's more of an informational doc ("how to contextual logging" doesn't seem to fit with the rest of the how-to titles, and also the content). Even though you've nested it under `use-logging`, imo it should still be a `how-to` type doc (e.g. [lxd docs](https://documentation.ubuntu.com/lxd/default/howto/).
> 
> Isn't this more like a reference or explanation?

Oh that is a good point, I never considered reading the section header that way.
I was debating between putting it in the explanation/logging subdirectory or under use-logging and settled on this because it does contain some descriptions of using/filtering the contextual logs.

What do you think about splitting this up, so that the reference part which explains how the contextual logging works is in the explanation section, but I add a small section providing examples of how to use a custom trace ID and filtering logs nested under the use-logging article?

tobiasdemendonca:
> > I have a general comment on this as I've been looking into the getting started docs.
> > I don't think this is a how-to doc, it's more of an informational doc ("how to contextual logging" doesn't seem to fit with the rest of the how-to titles, and also the content). Even though you've nested it under `use-logging`, imo it should still be a `how-to` type doc (e.g. [lxd docs](https://documentation.ubuntu.com/lxd/default/howto/).
> > Isn't this more like a reference or explanation?
> 
> Oh that is a good point, I never considered reading the section header that way. I was debating between putting it in the explanation/logging subdirectory or under use-logging and settled on this because it does contain some descriptions of using/filtering the contextual logs.
> 
> What do you think about splitting this up, so that the reference part which explains how the contextual logging works is in the explanation section, but I add a small section providing examples of how to use a custom trace ID and filtering logs nested under the use-logging article?

`explanation/logging` definitely seems like a good place for it. And adding those sections to `use-logging` seems reasonable too

tobiasdemendonca (approved):
A nice read!

PR #225, opened 2026-05-04
feat: add Temporal workflow to configure MAAS according to profile provided by Site Manager

wyattrees:


src/maascommon/workflows/msm.py
```diff
@@ -11,6 +11,7 @@
 MSM_RESTORE_DEFAULT_BOOT_SOURCE_WORKFLOW_NAME = (
     "msm-restore-default-boot-source"
 )
+MSM_CONFIGURE_PROFILE_WORKFLOW = "msm-configure-profile"
```
MaikRe:
```suggestion
MSM_CONFIGURE_PROFILE_WORKFLOW_NAME = "msm-configure-profile"
```

src/maastemporalworker/workflow/msm.py
```diff
@@ -25,10 +26,12 @@
 from maascommon.workflows.bootresource import MASTER_IMAGE_SYNC_WORKFLOW_NAME
 from maascommon.workflows.msm import (
     MachinesCountByStatus,
+    MSM_CONFIGURE_PROFILE_WORKFLOW,
```
MaikRe:
```suggestion
    MSM_CONFIGURE_PROFILE_WORKFLOW_NAME,
```

src/maastemporalworker/workflow/msm.py
```diff
@@ -724,3 +825,122 @@ async def run(self, input: MSMTokenRefreshParam) -> None:
                 ),
                 start_to_close_timeout=MSM_TIMEOUT,
             )
+
+
+@workflow.defn(name=MSM_CONFIGURE_PROFILE_WORKFLOW, sandboxed=False)
```
MaikRe:
```suggestion
@workflow.defn(name=MSM_CONFIGURE_PROFILE_WORKFLOW_NAME, sandboxed=False)
```

src/maastemporalworker/workflow/msm.py
```diff
@@ -724,3 +825,122 @@ async def run(self, input: MSMTokenRefreshParam) -> None:
...
+            "selections": workflow.start_activity(
+                MSM_SET_SELECTIONS_ACTIVITY_NAME,
+                MSMSetSelectionsParam(
+                    selections=full_config["selections"], sm_url=input.sm_url
+                ),
+                start_to_close_timeout=MSM_TIMEOUT,
+            ),
+        }
+        # need to handle trigger_image_sync separately becuase we only
+        # want to start the activity if it's specified as True
```
MaikRe: Spelling mistake in `because`, also I think the code is pretty self explanatory here, not sure if this comment is needed

src/maastemporalworker/workflow/msm.py
```diff
@@ -724,3 +825,122 @@ async def run(self, input: MSMTokenRefreshParam) -> None:
...
+        if full_config["trigger_image_sync"]:
```
MaikRe, line 902: Not sure if this could cause problems but what happens if the full config doesn't contain the needed key?
Applies for 'trigger_image_sync', `global_config` and `selections` in the lines above.
I think it would be good to explicitly check for their existence or use .get() to work with defaults if need be.

MaikRe (changes requested):
Some comments + nits inline, also copyright years for all four files need to be bumped

src/maastemporalworker/workflow/msm.py
wyattrees: This really should never happen, but I added a check just in case

MaikRe (approved):
lgtm

PR #226, opened 2026-05-04
docs: add docs for security logging

MaikRe:
Documentation for MA257


docs/explanation/security-logging.md
```diff
@@ -0,0 +1,463 @@
...
+
+### Key Fields
+
+- **message**: The security event type and optional additional information
+- **timestamp**: ISO 8601 formatted timestamp in UTC
+- **type**: Always set to `security` for security events
+- **trace_id**: Unique identifier correlating all logs for a single request
+- **userID**: Username of the authenticated user (when applicable)
+- **role**: User's role such as Admin or User (when applicable)
+- **token_hash**: SHA-256 hash of tokens for secure identification (for token events)
```
wyattrees: not present in the structure above?

docs/explanation/security-logging.md
```diff
@@ -0,0 +1,463 @@
...
+}
+```
+
+### Key Fields
+
+- **message**: The security event type and optional additional information
+- **timestamp**: ISO 8601 formatted timestamp in UTC
+- **type**: Always set to `security` for security events
+- **trace_id**: Unique identifier correlating all logs for a single request
+- **userID**: Username of the authenticated user (when applicable)
```
wyattrees, line 46: Why is this camel case while everything else is snake case?

docs/explanation/security-logging.md
```diff
@@ -0,0 +1,463 @@
...
+- **role**: User's role such as Admin or User (when applicable)
+- **token_hash**: SHA-256 hash of tokens for secure identification (for token events)
+
+## Security Event Types
+
+### Authentication Events
+
```
wyattrees, line 53: Maybe add some explanation here of what's to come in the subsections below?

alexsander-souza (dismissed):

docs/explanation/security-logging.md
```diff
@@ -0,0 +1,463 @@
...
+2. **Correlation**: The same token always produces the same hash, enabling tracking
+3. **Non-reversibility**: Hash cannot be reversed to obtain the original token
+4. **Auditability**: Enables security investigations without exposing credentials
+
+### Tracking Token Lifecycle
+
+You can track a token from creation through deletion and attempted reuse by matching `token_hash` values:
+
+```json
+// Token created
```
wyattrees: comments don't exist in JSON so I don't know how these would show up in a Markdown code block. Also, these comments are not really needed so I'd just remove them.

wyattrees (changes requested):
just a few minor things

docs/explanation/security-logging.md
MaikRe: They show up fine in the markdown block, but you are right that they are probably not needed so will remove.

docs/explanation/security-logging.md
MaikRe: The actual implementation is like this as well, but this is a good point. I will do a follow up PR that matches the userID to be snake case both here in the docs as well as in the code itself.

docs/explanation/security-logging.md
MaikRe: The token_hash is also a 'when applicable' item and it doesn't fit into the example structure I provide above.
I'll update the doc to say so

wyattrees (approved):
lgtm

PR #228, opened 2026-05-05
feat: start workflow to configure MAAS according to Site Manager profile from heartbeat workflow when configuration hashes do not match

wyattrees:


src/maastemporalworker/workflow/msm.py
```diff
@@ -104,6 +108,31 @@
...
+    return hashlib.sha256(serialized).hexdigest()
+
+
+def _normalize_config_value(obj: Any) -> Any:
+    """Recursively sort dict keys and sort all lists."""
+    if isinstance(obj, dict):
+        return {k: _normalize_config_value(obj[k]) for k in sorted(obj)}
+    if isinstance(obj, list):
+        return sorted(obj)
+    return obj
```
alemar99: I think it would be ideal to have a class for the configuration.
We can specify the possible attributes for the config together with these methods

src/maastemporalworker/workflow/msm.py
```diff
@@ -407,21 +436,43 @@ async def get_running_version(self) -> str:
...
+        """
+        async with self.start_transaction() as services:
+            cfg = await services.configurations.get_msm_config()
+            selections = await services.boot_source_selections.get_all_highest_priority()
+        full_normalized_config = _normalize_config_value(
+            {
+                "global_config": cfg,
+                "selections": [
+                    f"{s.os}/{s.release}/{s.arch}" for s in selections
+                ],
```
alemar99: we are sure that all the highest priority are from the MSM image source because we disable all the others right?

Could the user mess up this and enable another source other than MSM?

src/maastemporalworker/workflow/msm.py
```diff
@@ -407,21 +436,43 @@ async def get_running_version(self) -> str:
...
     async def get_known_config_options(self) -> list[str]:
         """Get the configuration changes that this MAAS knows
         how to handle.
 
         Returns:
             list[str]: the list of configuration option keys supported
         """
-        # TODO: return the actual options when they're implemented
-        return []
+        return list(CONFIGURATION_ACTIVITIES)
```
alemar99, line 483: Is it necessary to have an activity to return a constant? Or are we planning to do something here in the future and this is more a placeholder?

src/maastemporalworker/workflow/msm.py
```diff
@@ -747,30 +805,52 @@ async def run(self, input: MSMHeartbeatParam) -> None:
...
-                ),
             )
+            if current_hash != heartbeat_response.config_hash:
+                try:
+                    await workflow.start_child_workflow(
+                        MSM_CONFIGURE_PROFILE_WORKFLOW_NAME,
+                        MSMConfigureProfileParam(
+                            sm_url=input.sm_url, jwt=input.jwt
+                        ),
+                        id=f"{MSM_CONFIGURE_PROFILE_WORKFLOW_NAME}:{REGION_TASK_QUEUE}",
```
alemar99: Is the `REGION_TASK_QUEUE` necessary in the id? Did you mean to put this in the `task_queue` param?

alemar99 (changes requested):
Some questions inline

src/maastemporalworker/workflow/msm.py
wyattrees: do you mean a class for just `global_config` or the entire object, including `selections` and `trigger_image_sync`? If the former, I think that would be an unnecessarily large object. I'm following how we store the configurations in the database (only track what's set explicitly, everything not specified is the default).

If you mean the latter, sure I can create an object, but we'd still need this generic `_normalize_config_value`, since we need to make sure the values of the `global_config` are sorted if they are lists (or dicts, but I don't think that actually exists), and the type of this object's `global_config` would still be `dict[str, Any]`.

src/maastemporalworker/workflow/msm.py
wyattrees: Yes, all other image sources are disabled during enrollment. However, currently, there is nothing preventing users from adding additional sources (we may disable this in the future, though). I'd argue that it's still OK in this case--the selections don't necessarily _need_ to be from MSM here. As long as the image is available for download somewhere, we should let the user select it.

src/maastemporalworker/workflow/msm.py
wyattrees: This is more of a placeholder. When we eventually add multi-tenancy to MAAS, we also want the tenant groups to be consistent with what is defined in MSM, which may mean disabling certain (future) profile options.

src/maastemporalworker/workflow/msm.py
wyattrees: mean to put this in `task_queue`, will update

src/maastemporalworker/workflow/msm.py
alemar99: I was thinking at something like this:
```python
@dataclasses.dataclass
class MyConfig:
    global_config: dict[str, Any]
    selections: list[str]
    trigger_image_sync: bool

    def hash(self) -> str:
        ...

    def normalize(self):
        ...
```

src/maastemporalworker/workflow/msm.py
alemar99: Maybe we should prevent the user to enable other boot sources if msm is enabled then. I think that if you manage multiple MAAS instances you expect them to have the same version of noble for example, which might not be true if you don't check the boot source from which the image is downloaded

alemar99: ^ (not in this PR of course, something to keep in mind)

wyattrees: agreed, I'll add a ticket to do this in the near future

src/maastemporalworker/workflow/msm.py
wyattrees: done, let me know what you think

src/tests/maasservicelayer/services/test_configurations.py
```diff
@@ -424,3 +424,23 @@ async def test_clear_and_set_many_fails_secret(self):
...
+            context=Context(),
+            database_configurations_service=Mock(
+                DatabaseConfigurationsService
+            ),
+            secrets_service=Mock(SecretsService),
+            events_service=Mock(EventsService),
+        )
+        test_cfg = {
+            "theme": "dark",  # should be included
+            "kernel_ops": None,  # should not (default)
```
MaikRe:
```suggestion
            "kernel_opts": None,  # should not (default)
```
Spelling mistake or is this a different config item?

src/tests/maastemporalworker/workflow/test_msm.py
```diff
@@ -1013,7 +1073,95 @@ async def set_enrol(input: MSMConnectorParam) -> None:
...
+        assert len(calls["msm-send-heartbeat"]) == 2
+        assert len(calls["msm-get-enrol"]) == 2
+        assert len(calls["msm-get-version"]) == 2
+        assert len(calls["msm-get-config-options"]) == 1
+        mock_start_wf.assert_called_once_with(
+            MSM_RESTORE_DEFAULT_BOOT_SOURCE_WORKFLOW_NAME,
+            MSMRestoreDefaultBootSourceParam(
+                sm_url=_MSM_BASE_URL,
+            ),
+            id=f"{MSM_RESTORE_DEFAULT_BOOT_SOURCE_WORKFLOW_NAME}:{REGION_TASK_QUEUE}",
```
MaikRe, line 1244: This is different id formatting from the other tasks, which include the 'task_queue' parameter. Is that intentional?

src/maastemporalworker/workflow/msm.py
```diff
@@ -747,30 +814,53 @@ async def run(self, input: MSMHeartbeatParam) -> None:
...
-                    backoff_coefficient=1.0,
-                    initial_interval=timedelta(seconds=next_update),
-                ),
             )
+            if current_hash != heartbeat_response.config_hash:
+                try:
+                    await workflow.start_child_workflow(
+                        MSM_CONFIGURE_PROFILE_WORKFLOW_NAME,
+                        MSMConfigureProfileParam(
+                            sm_url=input.sm_url, jwt=input.jwt
```
MaikRe: Is using input.jwt correct here since a previous action pulls a new secret?
In the MSM_SEND_HEARTBEAT_ACTIVITY_NAME you use `jwt=secret["jwt"]

src/maastemporalworker/workflow/msm.py
```diff
@@ -104,6 +108,44 @@
...
+}
+
+
+def _normalize_config_value(obj: Any) -> Any:
+    """Recursively sort dict keys and sort all lists."""
+    if isinstance(obj, dict):
+        return {k: _normalize_config_value(obj[k]) for k in sorted(obj)}
+    if isinstance(obj, list):
+        return sorted(obj)
+    return obj
```
MaikRe, line 125: This sorts all config items that contain lists.
I don't know if this might be problematic for config items where ordering matters. 
For example for the `UpstreamDnsConfig` this sorting would provide the same hashes for ["8.8.8.8", "1.1.1.1"]  and ["1.1.1.1", "8.8.8.8"] but in practice the ordering determines which DNS services MAAS tries to use first.

src/tests/maasservicelayer/services/test_configurations.py
wyattrees: indeed a typo

src/maastemporalworker/workflow/msm.py
wyattrees: should come from the secret, will update.

src/tests/maastemporalworker/workflow/test_msm.py
wyattrees: This workflow was implemented a while ago and I'm trying to remember if there was a reason for the way this is called (i.e. with `REGION_TASK_QUEUE` as a part of the ID). It actually currently follows the way that child workflows are triggered elsewhere in this module, but I don't think it's actually necessary. Anyway, this is beyond the scope of this PR but could be refactored in the future.

src/maastemporalworker/workflow/msm.py
wyattrees: What is set from the profile isn't actually sorted. We don't sort items when we retrieve the full profile, only when we compare hashes

alemar99: I think we should add some unit tests around this

MaikRe: > What is set from the profile isn't actually sorted. We don't sort items when we retrieve the full profile, only when we compare hashes

If site manager is trying to set a config to [A,B] and MAAS has [B,A] the config update activity won't run since the hashes are the same, thus MAAS is never updated to [A,B]

wyattrees: okay yeah, I see how that's an issue. I'll add some exceptions for things where order matters.

@alemar99 unit tests around what exactly?

MaikRe (dismissed):
+1, but deferring to @alemar99  regarding the need for unit tests on this

src/maasservicelayer/models/configurations.py
```diff
@@ -1232,6 +1232,87 @@ class ConfigFactory:
         if config_model.is_public
     }
 
+    MSM_CONFIGS: dict[str, Type[Config]] = {
```
r00ta: Maintaining duplicated code is expensive and error prone. I would remove this and find another solution

r00ta (changes requested):

src/maasservicelayer/models/configurations.py
wyattrees: Added a `supported_by_msm` field to the base `Config` class

src/maastemporalworker/workflow/msm.py
alemar99: I was thinking about some unit tests to make sure the behavior is the expected one with nested objects

src/maastemporalworker/workflow/msm.py
```diff
@@ -104,6 +108,50 @@
...
+
+
+@dataclasses.dataclass
+class MSMConfigProfile:
+    global_config: dict[str, Any]
+    selections: list[str]
+    trigger_image_sync: bool
+
+    _RETAIN_ORDER_OPTIONS: ClassVar[list[str]] = ["upstream_dns"]
```
alemar99: Use the config name instead of hardcoding it?

alemar99 (approved):
LGTM

r00ta (approved):

PR #229, opened 2026-05-05
refactor: zones_with_summary to zones:statistics

nehjoshi:
- Refactored `/zones_with_summary` to `/zones:statistics`
- The UI will initially call `GET /zones` and for each `zone_id` retrieved, it will call `/zones:statistics?id=1&id=2&...`


src/maasapiserver/v3/api/public/handlers/zones.py
```diff
@@ -109,28 +105,32 @@ async def list_zones(
...
+                for zone_with_statistics in zones_with_statistics.items
             ],
-            total=zones_with_summary.total,
+            total=zones_with_statistics.total,
             next=(
-                f"{V3_API_PREFIX}/zones_with_summary?"
+                f"{V3_API_PREFIX}/zones:statistics?"
                 f"{pagination_params.to_next_href_format()}"
-                if zones_with_summary.has_next(
+                f"&{query_filters if query_filters else ''}"
```
alemar99: This will result in a link ending with `&` if `query_filters` is None.

src/maasapiserver/v3/api/public/handlers/zones.py
```diff
@@ -62,19 +62,15 @@ class ZonesHandler(Handler):
     async def list_zones(
         self,
         pagination_params: PaginationParams = Depends(),  # noqa: B008
-        filters: ZonesFiltersParams = Depends(),  # noqa: B008
```
alemar99, line 65: so you will not need to filter the `/zones` endpoint by id anymore?
Maybe we can leave it anyway?

alemar99 (changes requested):
A couple nits inline

src/maasapiserver/v3/api/public/handlers/zones.py
nehjoshi: Good catch, I should've shortened the last bracket!

src/maasapiserver/v3/api/public/handlers/zones.py
nehjoshi: Sure, I don't think it's used anywhere else, but I'll leave it in

nehjoshi:
Addressed prev. comments and added back the `test_list_with_filters` test that was previously removed

src/maasapiserver/v3/api/public/handlers/zones.py
```diff
@@ -61,8 +61,8 @@ class ZonesHandler(Handler):
     )
     async def list_zones(
         self,
-        pagination_params: PaginationParams = Depends(),  # noqa: B008
         filters: ZonesFiltersParams = Depends(),  # noqa: B008
+        pagination_params: PaginationParams = Depends(),  # noqa: B008
```
alemar99, line 64: Can you move this one line up as to not have any diff? Sorry for the nit

nehjoshi: Sure!

nehjoshi:
Reordered handler arguments

alemar99 (approved):
LGTM

PR #237, opened 2026-05-06
refactor: return proper error code if script result does not exist

MaikRe:
If script_sets.get(id) returns None the error regarding an unknown id should only be return if that id does not match the naming of one of the three known ids. (current-commissioning, current-testing, current-installation), otherwise the function should provide clear information that no script set exists.


r00ta (changes requested):
missing tests

r00ta (approved):

PR #238, opened 2026-05-07
feat: add filtering to custom_images endpoint based ID and file type

MaikRe:
Extends the CustomImageFilterParams to include file type, such that filtering for specific file types like 'self-extracting' is possible. Adds the CustomImageFilterParams to the custom_images GET endpoint. The primary use-case is getting a list of images to be used for switch provisioning based on the 'self-extracting' image type.

### DIFF
--- a/src/maasapiserver/v3/api/public/handlers/boot_resources.py
+++ b/src/maasapiserver/v3/api/public/handlers/boot_resources.py
@@ -329,7 +329,9 @@ class CustomImagesHandler(Handler):
                 ]
             )
             if filter_clause
-            else BootResourceClauseFactory.with_rtype(BootResourceType.UPLOADED)
+            else BootResourceClauseFactory.with_rtype(
+                BootResourceType.UPLOADED
+            )
         )
 
         boot_resources = await services.boot_resources.list(
--- a/src/maasapiserver/v3/api/public/handlers/boot_resources.py
+++ b/src/maasapiserver/v3/api/public/handlers/boot_resources.py
@@ -314,18 +314,41 @@ class CustomImagesHandler(Handler):
     )
     async def list_custom_images(
         self,
+        filters: CustomImageFilterParams = Depends(),  # noqa: B008
         pagination_params: PaginationParams = Depends(),  # noqa: B008
         services: ServiceCollectionV3 = Depends(services),  # noqa: B008
     ) -> ImageListResponse:
+        filter_clause = filters.to_clause()
+        where_clause = (
+            BootResourceClauseFactory.and_clauses(
+                [
+                    BootResourceClauseFactory.with_rtype(
+                        BootResourceType.UPLOADED
+                    ),
+                    filter_clause,
+                ]
+            )
+            if filter_clause
+            else BootResourceClauseFactory.with_rtype(BootResourceType.UPLOADED)
+        )
+
         boot_resources = await services.boot_resources.list(
             page=pagination_params.page,
             size=pagination_params.size,
-            query=QuerySpec(
-                where=BootResourceClauseFactory.with_rtype(
-                    BootResourceType.UPLOADED
-                )
-            ),
+            query=QuerySpec(where=where_clause),
         )
+
+        next_link = None
+        if boot_resources.has_next(
+            pagination_params.page, pagination_params.size
+        ):
+            next_link = (
+                f"{V3_API_PREFIX}/custom_images?"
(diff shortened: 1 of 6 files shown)


ndv99 (commented):
Tentative approval as a UI person 😅 we'll wait for Alessandro to get back for a proper review

src/tests/maasservicelayer/db/repositories/test_bootresources.py
```diff
@@ -101,6 +102,22 @@ def test_with_selection_boot_source_id(self) -> None:
...
+        compiled = str(
+            clause.condition.compile(compile_kwargs={"literal_binds": True})
+        )
+        assert "maasserver_bootresource.id IN" in compiled
+        assert "SELECT DISTINCT" in compiled
+        assert "maasserver_bootresourceset.resource_id" in compiled
+        assert (
+            "maasserver_bootresourcefile.filetype = 'self-extracting'"
+            in compiled
+        )
```
alemar99: I'd prefer to match the complete sql query

src/tests/maasservicelayer/db/repositories/test_bootresources.py
```diff
@@ -341,3 +358,247 @@ async def test_list_custom_image_statistics(
...
+        fixture: Fixture,
+    ):
+        """Test filtering boot resources by file type."""
+        from maasservicelayer.db.filters import QuerySpec
+        from tests.fixtures.factories.bootresourcefiles import (
+            create_test_bootresourcefile_entry,
+        )
+        from tests.fixtures.factories.bootresourcesets import (
+            create_test_bootresourceset_entry,
+        )
```
alemar99: Can't we import this at the top of the file?

src/tests/maasservicelayer/db/repositories/test_bootresources.py
```diff
@@ -341,3 +358,247 @@ async def test_list_custom_image_statistics(
...
+            size=2048,
+            filename_on_disk="installer.bin",
+            resource_set_id=switch_set.id,
+        )
+
+        # Create boot resource with tgz file (regular image)
+        regular_resource = await create_test_bootresource_entry(
+            fixture,
+            rtype=BootResourceType.UPLOADED,
+            name="ubuntu/focal",
```
alemar99: nit: we don't support uploading a custom image with name "ubuntu"

src/maasapiserver/v3/api/public/models/requests/boot_resources.py
```diff
@@ -309,13 +309,33 @@ class CustomImageFilterParams(BaseModel):
             description="Filter by Custom Image ID",
         )
     )
+    file_type: BootResourceFileType | None = Field(
```
alemar99: Use `BootResourceFileTypeChoice` for a more user friendly value? We never exposed this to the API, so I think it's better to be consistent and use the other one

src/maasapiserver/v3/api/public/handlers/boot_resources.py
```diff
@@ -314,18 +314,43 @@ async def upload_custom_image(
...
+                        BootResourceType.UPLOADED
+                    ),
+                    filter_clause,
+                ]
+            )
+            if filter_clause
+            else BootResourceClauseFactory.with_rtype(
+                BootResourceType.UPLOADED
+            )
+        )
```
alemar99: you can also pass a single clause to `and_clauses` so you can write this in a more concise way.

[SQA docs](https://docs.sqlalchemy.org/en/21/core/sqlelement.html#sqlalchemy.sql.expression.and_)

alemar99 (changes requested):
Some nits inline

src/maasapiserver/v3/api/public/models/requests/boot_resources.py
MaikRe: ~~I don't think this is the right thing since the choice item doesn't contain the prefix of "ROOT" and thus doesn't end up matching the actual database contents in the filter clause that's constructed.~~

src/maasapiserver/v3/api/public/handlers/boot_resources.py
MaikRe: I am struggling to simplify this in a way that seems 'pretty' to me, as the filter_clause may be none and passing that to the `and_clauses` would not considered valid, so some kind of conditional catching of that case is required. Can you specify what you mean here?

alemar99 (approved):
+1

PR #239, opened 2026-05-07
refactor: deduplicate Script lookup by name or id in scripts API

tanzin8r:
A small cleanup in the scripts API that removes repeated logic for looking up a script by numeric id or by name. A single helper in NodeScriptHandler now does that lookup, and read, delete, update, download, revert, add_tag, and remove_tag all call that helper instead of repeating the same conditional block in each method. I didn't add tests because existing TestScriptAPI coverage already runs these endpoints in both by name and by id scenarios.


wyattrees (approved):
+1

PR #240, opened 2026-05-07
feat: add CLI and API V2 implementation for switch management

MaikRe:


src/maasserver/api/switches.py
```diff
@@ -0,0 +1,308 @@
...
+        Boot resource ID if image name is provided and found, None otherwise
+
+    Raises:
+        MAASAPIValidationError: If image name is provided but not found, or if
+            short format name doesn't reference an ONIE image
+    """
+    try:
+        return service_layer.exec_async(
+            resolve_image_id(
+                image, service_layer.services.service_collection
```
MaikRe: I created this PR to track some initial work on bringing switch management to V2/CLI.
I am unsure I am using the async adapter correctly here.
The problem is that the resolve_image_id function expects the service_collection as a parameter, and currently this is only available via `service_layer.services._service_collection`. 
I have added another property that exposes this in the sqlalchemy file, but in order for ruff to not complain about it it has required changing the type of the existing `services` property.
@r00ta maybe you can provide some advice on this.

src/maasserver/sqlalchemy.py
```diff
@@ -121,7 +126,7 @@ def init(self):
     # technically not the right type. However, it's fine because this adapter is supposed to be used only in the django
     # application and we will never run this process with mypy.
     @property
-    def services(self) -> ServiceCollectionV3:
```
r00ta, line 124: keep as before, look at the comment above

MaikRe:
Closing as switch management will only be an experimental feature in 3.8 and limited to API V3.

PR #241, opened 2026-05-07
fix: dedupe script results when filters match the same row twice

tanzin8r:
filter_script_results appended once per matching filter token, so a single script result could appear multiple times when several tokens matched it for example same name twice, or name plus tag. We should stop after the first match per script result so combined filters stay. Also added a regression test that passes both script name and a tag that apply to the same result.


wyattrees (approved):
+1

PR #243, opened 2026-05-11
fix: pin twisted version in docs reqs

AloizioMacedo:
[Twisted 26.0.4](https://github.com/twisted/twisted/releases/tag/twisted-26.4.0) breaks our CI.


abuyukyi101198 (approved):
LGTM

PR #244, opened 2026-05-11
fix: use release name when uploaded image title is empty

tanzin8r:
Uploaded custom images can store an empty string in extra["title"]; Changing that to treat it like a missing title so deploy/release labels fall back to the release name instead of a blank string. [LP:2044378](https://bugs.launchpad.net/maas/+bug/2044378)


src/maasserver/utils/osystems.py
```diff
@@ -108,9 +108,13 @@ def list_all_usable_osystems() -> Dict[str, OperatingSystem]:
...
-            # User may set the title of an uploaded resource.
-            if "title" in br.extra:
-                release_title = br.extra["title"]
+            # User may set the title of an uploaded resource. Empty values
+            # in extra must not override the display name (LP:2044378).
+            explicit_title = br.extra.get("title")
+            if isinstance(explicit_title, str):
+                explicit_title = explicit_title.strip()
+            if explicit_title:
+                release_title = explicit_title
```
alemar99:
```suggestion
            if (title := br.extra.get("title")) and title.strip() != "":
                release_title = title.strip()
```

alemar99 (changes requested):
See Inline

src/maasserver/utils/osystems.py
```diff
@@ -108,9 +108,13 @@ def list_all_usable_osystems() -> Dict[str, OperatingSystem]:
...
             else:
                 release_title = release_name
```
AloizioMacedo:
```suggestion
            explicit_title = br.extra.get("title", "").strip()
            release_title = explicit_title if explicit_title else release_name
```

---

Some comments: using `isinstance` kind of suggests that `explicit_title` could be something like int or something else, when it actually looks like it could only be None. It would be better to check explicitly for None instead.

But even then, this has a lot of explicit branching which makes it a little confusing to read. 

The suggestion above reads more easily (e.g., less assignments) and it also removes indentations.

AloizioMacedo:
It isn't clear to me if this is really an issue in the API, or rather in the UI. 

Are we deliberately allowing custom images to be created without a title, or is that a bug? If we are, then my intuition is that this fix should be done in the UI side. If we are not, maybe the fix is then to make it required?

I would talk to the UI team before merging this, and also check the business logic and assumptions around custom images.

tanzin8r:
> It isn't clear to me if this is really an issue in the API, or rather in the UI.
> 
> Are we deliberately allowing custom images to be created without a title, or is that a bug? If we are, then my intuition is that this fix should be done in the UI side. If we are not, maybe the fix is then to make it required?
> 
> I would talk to the UI team before merging this, and also check the business logic and assumptions around custom images.

it has always been optional. I checked and the CLI and API docs both say title is optional for boot-resources create. I don't think we should make it required now. Basically  the original code was storing `"" in extra["title"] ` instead of omitting the key entirely, so the presence check "title" in br.extra passed and wiped out the fallback.  

The UI just renders whatever label the region hands it. Looping in @abuyukyi101198 I put up a fix for the bug I was mentioning yesterday. There's also a second symptom in the bug comments: on the machines list, a deployed machine that used a custom image shows only "Deployed" instead of the image/OS name. I  checked that but it seems to be working as well with this fix. 

However there is one issue worth noting during my testing of the fix. when i am uploading a new custom image, the deploy dropdown doesn't show it until I refresh the page. I suspect this is because the dropdown gets its list of OS/releases from a WebSocket call (general.osinfo) that only fires once when the page loads. There is no push notification when a boot resource changes, so the UI never knows to update itself.

Every other update in MAAS gets a Postgres notification when something changes and the UI reacts automatically. Boot resources don't have that the UI just holds on to whatever it fetched at page load.

src/maasserver/utils/osystems.py
tanzin8r: Thanks! I simplified it along these lines. I think using get("title", "") would be an issue only if someone patches the DB directly, which the bug report shows as the workaround.
 (get("title") or "") handles both the missing key and the None case.
`explicit_title = (br.extra.get("title") or "").strip() `
`release_title = explicit_title or release_name`

alemar99 (approved):
+1

r00ta:
why was this closed?

PR #245, opened 2026-05-12
fix: explicitly configure dhcp in ip= kernel command line

enr0n:
The ip= kernel command line option is implemented slightly differently across dracut, initramfs-tools, systemd-network-generator, etc.

Dracut now lets systemd-network-generator handle ip= parsing, and systemd-network-generator requires that the DHCP type is set when the interface name is. E.g.,

 ip=::::bibarel:BOOTIF:dhcp BOOTIF=01-ec:b1:d7:7f:2a:6c

is valid but

 ip=::::bibarel:BOOTIF BOOTIF=01-ec:b1:d7:7f:2a:6c

is not. Ubuntu currently carries a downstream patch to allow for easier migration from initramfs-tools, but going forward the DHCP type should be set explicitly. Hence, set it to "dhcp", which is the behavior when unspecified in initramfs-tools.

Resolves: LP#2141588

### DIFF
--- a/src/metadataserver/builtin_scripts/tests/test_hooks.py
+++ b/src/metadataserver/builtin_scripts/tests/test_hooks.py
@@ -1238,7 +1238,7 @@ KERNEL_CMDLINE_OUTPUT = (
     "BOOT_IMAGE=http://10.245.136.6:5248/images/ubuntu/amd64/generic/bionic/"
     "daily/boot-kernel nomodeset ro root=squash:http://10.245.136.6:5248/"
     "images/ubuntu/amd64/generic/bionic/daily/squashfs "
-    "ip=::::flying-marmot:BOOTIF ip6=off overlayroot=tmpfs "
+    "ip=::::flying-marmot:BOOTIF:dhcp ip6=off overlayroot=tmpfs "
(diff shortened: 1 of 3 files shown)


r00ta (commented):
@enr0n is this backwards compatible? What if the user is deploying another previous version of Ubuntu?

enr0n:
> @enr0n is this backwards compatible? What if the user is deploying another previous version of Ubuntu?

I *believe* so, but it has been a little while since I checked (and I sometimes confuse the various implementations), so I will double check.

enr0n:
> @enr0n is this backwards compatible? What if the user is deploying another previous version of Ubuntu?

Sorry for the delay, I was OOO for a bit. Yes, this is backwards compatible. See e.g. this test case for initramfs-tools: https://git.launchpad.net/ubuntu/+source/initramfs-tools/tree/debian/tests/qemu-net#n81.

I.e. it is supported to explicitly configure "dhcp" or so in that field.

enr0n:
@r00ta gentle ping. I pushed with my canonical email to fix the CLA check.

r00ta (approved):
lgtm

PR #246, opened 2026-05-12
feat(api): add username_or_email filter to user list endpoint

ndv99:
- added `username_or_email` parameter to base user list endpoint
- created test for this parameter
- serialized ids parameter in `to_href_format`


alemar99 (dismissed):
LGTM

alemar99 (approved):

PR #256, opened 2026-05-14
POC: MCP server

alexsander-souza:


r00ta (changes requested):
Many models have been duplicated in the MCP server, this is going to add additional maintenance work. If we add this to master and we distribute it with MAAS itself, it means we will have to maintain it as part of the product. If that's the case, we must  not add this as a slop. We need to add TLS, ensure this is production ready and so forth and so on. 

Other options could be to distribute this as a separated snap (maybe community-maintained) that can be plugged to MAAS.

PR #257, opened 2026-05-14
refactor: remove custom dhcp and dqlite

troyanov:


troyanov:
Existing custom DHCP implementation is not fully-functional and currnetly there are no plans to continue working on it. Because Microcluster was a mandatory dependency for a home-made DHCP server, this PR also removes it.

alexsander-souza (approved):
Image

PR #258, opened 2026-05-14
chore: add python3 to the staging packages in snapcraft.yaml

r00ta:
Core base snaps up to and including core24 bundle the Python interpreter alongside a selection of Python libraries. But from core26 onwards, Python packages is no longer be included so we have to add it to the staging packages manually.


nehjoshi (approved):
LGTM

PR #261, opened 2026-05-14
fix: allow for withdrawal from MSM multiple times

wyattrees:
Previously, withdrawing from MSM multiple times would fail to restore the default boot source.


maas-lander:
- [x] Direct message
- [x] ~

skyecas (approved):

PR #262, opened 2026-05-14
chore: add workshop definition

bjornt:
This adds a dev workshop which works well enough so that you can unit tests inside it.

A github workflow was added, which can be triggered manually to confirm that it works as it should.


maas-lander:
- [ ] Direct message
- [ ] ~maas

MaikRe (approved):
lgtm!

PR #263, opened 2026-05-14
refactor: clean up docs before building snap

MaikRe:


maas-lander:
- [ ] Direct message
- [x] ~maas

AloizioMacedo (approved):

PR #265, opened 2026-05-15
chore: include python3 in snap staging packages

r00ta:


maas-lander:
- [ ] Direct message
- [ ] ~maas

abuyukyi101198 (approved):

PR #266, opened 2026-05-19
fix: prevent concurrent scale-up from over-allocating RPC connections

r00ta:
The previous implementatino of `scale_up_connections` was not taking into account that new connections have to go through a long-complex asynchronous process before becoming available. This lead concurrent calls to scale up connections without limit, because all the calls were evaluating `len(conns) < max_connections` as true, without taking into consideration the pending connections.   

With this change, `scale_up_connections` tracks in-flight connections with a pending counter and wait for these connections to become ready (with a 30s timeout) before releasing the slot. 

In addition to that, `_reap_extra_connection` is now called with a timeout `_keepalive / 1000` because `keepalive` is in milliseconds, but `callLater` expects seconds. 

Resolves: LP:2074122


maas-lander:
- [ ] Direct message
- [ ] ~maas

alemar99 (dismissed):
Great finding, LGTM!

nehjoshi (commented):
I think someone with more experience with this code than me can provide a better review, but I did have a question:
The ` _reap_extra_connection()` function calls `self.clock.callLater(self._keepalive, ...)` but here we are not dividing the `_keepalive` by 1000. Shouldn't it divide by a thousand as `callLater` expects seconds?

src/provisioningserver/rpc/tests/test_connectionpool.py
```diff
@@ -343,3 +348,66 @@ def test_remove_connection_removes_connection_from_pool(self):
...
+        disconnect = self.patch(cp, "disconnect")
+
+        d1 = cp.scale_up_connections()
+        self.assertEqual(cp._pending_connections[eventloop], 1)
+
+        ready_deferred.errback(CancelledError())
+
+        self.assertRaises(exceptions.MaxConnectionsOpen, extract_result, d1)
+        self.assertNotIn(eventloop, cp._pending_connections)
+        disconnect.assert_called_once_with(new_conn)
```
nehjoshi, line 413: The `test__reap_extra_connection_waits_for_a_busy_connection()` test in this file (not visible in the diff) will also need to be updated so that
```py
self.assertEqual(cp._keepalive / 1000, clock.calls[0].time)
```

nehjoshi (changes requested):
One last comment ->

nehjoshi (approved):
Thanks for addressing, LGTM!

PR #268, opened 2026-05-19
fix: handle conflicts when creating the default selection in the migration

alemar99:


maas-lander:
- [ ] Direct message
- [ ] ~maas

r00ta (approved):

PR #269, opened 2026-05-20
fix: prevent concurrent scale-up from over-allocating RPC connections

r00ta:
The previous implementatino of `scale_up_connections` was not taking
into account that new connections have to go through a long-complex
asynchronous process before becoming available. This lead concurrent
calls to scale up connections without limit, because all the calls were
evaluating `len(conns) < max_connections` as true, without taking into
consideration the pending connections.

With this change, `scale_up_connections` tracks in-flight connections
with a pending counter and wait for these connections to become ready
(with a 30s timeout) before releasing the slot.

In addition to that, `_reap_extra_connection` is now called with a
timeout `_keepalive / 1000` because `keepalive` is in milliseconds, but
`callLater` expects seconds.

Resolves: LP:2074122

(cherry-picked from 6fc05cb47935ea20a1bdec90d1825a13af2389e8)

### DIFF
--- a/src/provisioningserver/rpc/clusterservice.py
+++ b/src/provisioningserver/rpc/clusterservice.py
@@ -934,9 +934,16 @@ class ClusterClientService(TimerService):
         except exceptions.NoConnectionsAvailable:
             return self._tryUpdate().addCallback(call, self.getClient)
         except exceptions.AllConnectionsBusy:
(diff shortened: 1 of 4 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

PR #270, opened 2026-05-20
fix: stop Temporal sync-selection workflow when deleting selection

alemar99:
Resolves: LP:2152210


maas-lander:
- [x] Direct message
- [ ] ~maas

alexsander-souza (approved):

PR #271, opened 2026-05-20
refactor: relax simplestreams model validation

alemar99:
Default optional fields to None, otherwise Pydantic will raise a validation error if the field is not present.

(cherry picked from commit ccd1496c0268152c3e56a1bd1bdb80814b408bbb)


maas-lander:
- [ ] Direct message
- [ ] ~maas

PR #272, opened 2026-05-20
fix(network): don't convert STICKY IPs to DISCOVERED if machine receives same IP LP#2152681

nehjoshi:
Fix for [LP#2152681](https://bugs.launchpad.net/maas/+bug/2152681)

- Bug states that sometimes, STICKY IPs are converted to DISCOVERED when a machine is re-commissioned
- This is an issue because it then blocks re-assignment of the same (previously-STICKY but now DISCOVERED) IP
- The (not-so-ideal) workaround is to just manually manually release that IP via the CLI
- The issue stems from the `configure-dhcp` workflow not being correctly called due to the wrong conditions in an `if(...)` clause.
- When a machine is commissioned, it's network configuration is cleared. However, due to that faulty "if" block mentioned above, the DHCP configuration was not being properly updated, which resulted in old IPs still sticking around. Then, when a machine receives that same IP, it is converted to DISCOVERED, causing the bug the user mentioned.

### DIFF
--- a/src/maasserver/models/tests/test_dnsresource.py
+++ b/src/maasserver/models/tests/test_dnsresource.py
@@ -12,7 +12,7 @@ from maasserver.models.dnsresource import DNSResource, separate_fqdn
 from maasserver.permissions import NodePermission
 from maasserver.testing.factory import factory
 from maasserver.testing.testcase import MAASServerTestCase
-from maasserver.utils.orm import reload_object
+from maasserver.utils.orm import post_commit_hooks, reload_object
 
 
 class TestDNSResourceManagerGetDNSResourceOr404(MAASServerTestCase):
@@ -279,7 +279,8 @@ class TestDNSResource(MAASServerTestCase):
             alloc_type=IPADDRESS_TYPE.USER_RESERVED
         )
         dnsresource.ip_addresses.add(sip2)
-        dnsresource.delete()
+        with post_commit_hooks:
+            dnsresource.delete()
         assert StaticIPAddress.objects.filter(ip=sip1.get_ip()).exists()
         assert not StaticIPAddress.objects.filter(ip=sip2.get_ip()).exists()
 
@@ -314,7 +315,8 @@ class TestStaticIPAddressSignals(MAASServerTestCase):
     def test_non_orphaned_record_not_deleted(self):
         dnsrr = factory.make_DNSResource(ip_addresses=["8.8.8.8", "8.8.4.4"])
         sip = StaticIPAddress.objects.get(ip="8.8.4.4")
-        sip.delete()
+        with post_commit_hooks:
+            sip.delete()
         dnsrr = reload_object(dnsrr)
         sip = StaticIPAddress.objects.get(ip="8.8.8.8")
         self.assertIn(sip, dnsrr.ip_addresses.all())
--- a/src/maasserver/models/tests/test_staticipaddress.py
+++ b/src/maasserver/models/tests/test_staticipaddress.py
@@ -1725,7 +1725,7 @@ class TestStaticIPAddress(MAASServerTestCase):
             alloc_type=IPADDRESS_TYPE.AUTO,
             ip="10.0.0.1",
             subnet=subnet,
-            temp_expires_on=timezone.now(),
+            temp_expires_on=None,
         )
 
         with post_commit_hooks:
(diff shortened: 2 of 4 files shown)


maas-lander:
- [x] Direct message
- [ ] ~maas

r00ta (changes requested):
In general it's good that MAAS prevents the user to assign addresses that were previously discovered. This is preventing the user to end up with potentially duplicated IPs on the network. 

The only legit case is when they want to turn a discovered ip into a static ip for the same interface. 

When the user recommissions the machine, MAAS will delete all the IP addresses. 
I believe the current bug occurs when dhcp does not get updated in time and sends out the same IP, which is then detected as discovered by MAAS - this aspect might be legit for a bug report. If the user wants to retain the same network config they can select "skip networking" when they recommission the machine.

tobiasdemendonca:
I have been looking at this too, a fair chunk of this is new to me but I so far agree with Jacopo, and I think this is caused by incorrect copying over of trigger logic. See my comment here https://bugs.launchpad.net/maas/+bug/2152681

r00ta:
> I have been looking at this too, a fair chunk of this is new to me but I so far agree with Jacopo, and I think this is caused by incorrect copying over of trigger logic. See my comment here https://bugs.launchpad.net/maas/+bug/2152681

You're right, it seems like we don't trigger the DHCP update at all. Looking at https://github.com/canonical/maas/blame/726f481692edf2abf7e6e2043dcacc202cd49ba0/src/maasserver/triggers/system.py#L852 we have to update the logic in the django model

src/maasserver/models/tests/test_staticipaddress.py
nehjoshi, line 1780: Surprising how this test was meant to check that the workflow is **not** called, yet it was previously asserting that the workflow is triggered once  - the complete opposite

src/maasserver/models/tests/test_staticipaddress.py
```diff
@@ -1761,17 +1783,16 @@ def test_delete_empty_ip_does_not_call_dhcp_configure_workflow(self):
         )
         subnet = factory.make_Subnet(cidr="10.0.0.0/24")
         ip = StaticIPAddress(
-            alloc_type=IPADDRESS_TYPE.USER_RESERVED,
-            ip="10.0.0.1",
+            alloc_type=IPADDRESS_TYPE.STICKY,
```
nehjoshi, line 1786: `USER_RESERVED` type IPs cannot be set to `None`, hence I have changed it to sticky

src/maasserver/models/tests/test_dnsresource.py
```diff
@@ -314,7 +315,8 @@ def test_deletes_orphaned_record(self):
     def test_non_orphaned_record_not_deleted(self):
         dnsrr = factory.make_DNSResource(ip_addresses=["8.8.8.8", "8.8.4.4"])
         sip = StaticIPAddress.objects.get(ip="8.8.4.4")
-        sip.delete()
+        with post_commit_hooks:
```
tobiasdemendonca, line 318: Not too familiar with this, why do we need these now out of curiosity?

nehjoshi: Without it, we will get this nasty-looking error:
```sh
AssertionError: One or more post-commit tasks were present at the end of this test:
== Hook 1: <Deferred at 0x7250380ddac0> ==
 +- callback: <function callOut at 0x7250436ddf30>
 |      args: (<MagicMock id='125688863901408'>,)
 |    kwargs: {'workflow_name': 'configure-dhcp', 'param': ConfigureDHCPParam(system_ids=None, vlan_ids=None, subnet_ids=[None], static_ip_addr_ids=None, ip_range_ids=None, reserved_ip_ids=None), 'task_queue': 'region'}
 |   errback: <function _failthru at 0x725045800930>
```

tobiasdemendonca: Ahh yep, because the dhcp workflow now runs we need to to clean up post commit hooks, and you're doing that on all static ip deletions :+1: ignore me

tobiasdemendonca (approved):
LGTM from my end. That said, it may be worth getting a second pair of eyes from someone with more experience

src/maasserver/models/tests/test_staticipaddress.py
r00ta: Yeah, I'm not surprised actually

r00ta (approved):

PR #276, opened 2026-05-21
docs: add explicit configuration commands

AloizioMacedo:
Those commands were with ellipsis due to multiple ones existing under the same prefix. We are essentially just expanding them.


maas-lander:
- [ ] Direct message
- [ ] ~maas

AloizioMacedo:
For reference, this is how they look after the change:
image

tobiasdemendonca (approved):
LGTM

PR #279, opened 2026-05-27
fix: candid/rbac integration for 3.8

alemar99:
Since we deleted the django session id authentication provider, the v3 API is no longer accessible in 3.8+ using the old mechanism. This commit enables the macaroon-based authentication by setting the `macaroon-maas` cookie when a successful discharge request is done. In this way, the UI can use the macaroon to access the v3 API endpoints.

Other fixes:
- Macaroons: clear all cookies on discharge required, forcing a re-login through the UI
- /v3/auth/logout now clears all types of possible cookies
- V3: add all known cookie names as Enums
- V3: add a `revoke_refresh_token` for local JWT tokens


maas-lander:
- [ ] Direct message
- [ ] ~maas

src/maasapiserver/v3/auth/cookie_manager.py
```diff
@@ -21,6 +21,15 @@
 MAAS_OAUTH2_REFRESH_TOKEN_COOKIE_NAME = "maas.oauth2_refresh_token_cookie"
 
 
+class MAASDjangoCookie(StrEnum):
+    SESSION_ID = "sessionid"
+    CSRF_TOKEN = "csrf-token"
```
nehjoshi: This should be `csrftoken`

src/maasservicelayer/services/tokens.py
```diff
@@ -66,7 +79,23 @@ async def create_refresh_token(
...
+        logger.info(
+            f"{AUTHN_TOKEN_CREATED}:JWT:{REFRESH_TOKEN}",
+            type=SECURITY,
+            token_hash=hash_token_for_logging(token),
+        )
+        return refresh_token
+
+    async def revoke_refresh_token(self, token: str) -> None:
+        await super().delete_one(
+            query=QuerySpec(where=RefreshTokenClauseFactory.with_token(token))
```
nehjoshi: Shouldn't we be deleting the hashed token, so that:
```py
token_hash = hashlib.sha256(token.encode()).hexdigest()
await super().delete_one(
   query=QuerySpec(where=RefreshTokenClauseFactory.with_token(token_hash)
)
```
? Because we also save the hashed token in the DB

nehjoshi (changes requested):
Some comments inline

src/maasapiserver/v3/auth/cookie_manager.py
alemar99: thanks!

src/maasservicelayer/services/tokens.py
alemar99: Good finding!

nehjoshi (approved):
LGTM!

PR #280, opened 2026-05-27
fix(network): relax constraints for resizing dynamic range reservations

bryanfraschetti:
When an IP is allocated inside a dynamic range reservation, the validation performed during a resize sees the range as discontinuous, since an allocated IP splits the available space into separate contiguous blocks. As a result, both shrinking and expanding the reservation are rejected with "Requested dynamic range conflicts with an existing IP address or range", even when the requested change is otherwise valid. 

This commit addresses the issue by comparing the requested range against the previously persisted range and requiring only the newly added segments (i.e., the expanded portions) to be contained within an unused range. Pure shrinks add no segments and are always allowed. 

Resolves LP:2143090

### DIFF
--- a/src/maasserver/models/iprange.py
+++ b/src/maasserver/models/iprange.py
@@ -97,6 +97,12 @@ class IPRange(CleanSave, TimestampedModel):
     """Represents a range of IP addresses used for a particular purpose in
     MAAS, such as a DHCP range or a range of reserved addresses."""
 
+    @classmethod
+    def from_db(cls, db, field_names, values):
+        instance = super().from_db(db, field_names, values)
+        instance._loaded_values = dict(zip(field_names, values))
+        return instance
+
     objects = IPRangeManager()
 
     subnet = ForeignKey(
@@ -296,17 +302,20 @@ class IPRange(CleanSave, TimestampedModel):
         if self.id is None or self.type != IPRANGE_TYPE.DYNAMIC:
             return False
 
-        existing = (
-            type(self)
-            .objects.filter(id=self.id)
-            .values("type", "start_ip", "end_ip")
-            .first()
-        )
-        if existing is None or existing["type"] != IPRANGE_TYPE.DYNAMIC:
+        loaded_values = getattr(self, "_loaded_values", None)
+        if not loaded_values:
             return False
 
-        existing_start_ip = IPAddress(existing["start_ip"])
-        existing_end_ip = IPAddress(existing["end_ip"])
+        if loaded_values.get("subnet_id") != self.subnet_id:
+            return False
+
+        existing_start_ip = loaded_values.get("start_ip")
+        existing_end_ip = loaded_values.get("end_ip")
+        if existing_start_ip is None or existing_end_ip is None:
+            return False
+
+        existing_start_ip = IPAddress(existing_start_ip)
+        existing_end_ip = IPAddress(existing_end_ip)
 
         start = int(start_ip)
         end = int(end_ip)
@@ -338,8 +347,16 @@ class IPRange(CleanSave, TimestampedModel):
 
         return True
 
+    def _cache_loaded_values(self):
+        self._loaded_values = {
+            "subnet_id": self.subnet_id,
+            "start_ip": self.start_ip,
+            "end_ip": self.end_ip,
+        }
+
     def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
+        self._cache_loaded_values()
 
         if self.subnet.vlan.dhcp_on:
             post_commit_do(
--- a/src/maasserver/models/iprange.py
+++ b/src/maasserver/models/iprange.py
@@ -274,15 +274,69 @@ class IPRange(CleanSave, TimestampedModel):
         else:
             message += "IP address or range."
 
-        # Find unused range for start_ip
+        start_ip = IPAddress(self.start_ip)
+        end_ip = IPAddress(self.end_ip)
+
+        # Success when start and end are in the same unused range.
         for unused_range in unused:
-            if IPAddress(self.start_ip) in unused_range:
-                if IPAddress(self.end_ip) in unused_range:
-                    # Success, start and end IP are in an unused range.
-                    return
-                else:
-                    self._raise_validation_error(message)
+            if start_ip in unused_range and end_ip in unused_range:
+                return
+
+        if self._is_existing_dynamic_range_resize_allowed(
+            start_ip, end_ip, unused
+        ):
+            return
+
         self._raise_validation_error(message)
+
+    def _is_existing_dynamic_range_resize_allowed(
+        self, start_ip: IPAddress, end_ip: IPAddress, unused
+    ) -> bool:
+        """Allow resizing a dynamic range despite existing in-range allocations."""
+        if self.id is None or self.type != IPRANGE_TYPE.DYNAMIC:
+            return False
+
+        existing = (
+            type(self)
+            .objects.filter(id=self.id)
+            .values("type", "start_ip", "end_ip")
+            .first()
+        )
+        if existing is None or existing["type"] != IPRANGE_TYPE.DYNAMIC:
+            return False
+
+        existing_start_ip = IPAddress(existing["start_ip"])
+        existing_end_ip = IPAddress(existing["end_ip"])
+
+        start = int(start_ip)
+        end = int(end_ip)
+        existing_start = int(existing_start_ip)
+        existing_end = int(existing_end_ip)
+
+        added_segments = []
+
+        left_end = min(end, existing_start - 1)
+        if start <= left_end:
+            added_segments.append((start, left_end))
+
+        right_start = max(start, existing_end + 1)
+        if right_start <= end:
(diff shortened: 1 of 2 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

r00ta (changes requested):
say you have 
- a dynamic range 172.16.0.10-172.16.0.20 
- an already allocated AUTO IP 172.16.0.25 

and you want to resize to 172.16.0.10-172.16.0.30. Is your change going to allow the DHCP server to send out 172.16.0.25 as well?

tanzin8r:
Thanks for the fix.

### Implementation

- Consider caching the original `start`, `end`, and `subnet` when the model is loaded from the database using `from_db`, so validation does not need an extra query on every `clean()`.
- The same overlap check exists in the v3 `ipranges` API. I can help with that part if you want.
- When the range is moved to another subnet, the resize exception should not apply. Use full validation on the new subnet instead.

### Testing

I ran additional cases on top of the PR tests, including the subnet-change guard.

**Expected to pass on resize:**

1. **Shrink end with lease inside pool**  
   Range `10–50`, lease at `.30`, end changed to `.49`.  
   Expected: `clean()` and `save()` succeed. Only addresses at the tail are removed; nothing new is claimed.

2. **Expand end with lease inside pool**  
   Range `10–50`, lease at `.30`, end changed to `.60`.  
   Expected: succeed. Only `51–60` is new and should be free. The lease inside the old range should not block this.  
   I also checked on master without the fix: this fails with error from the bug report. With the PR it passes.

3. **Shrink start with lease inside pool**  
   Range `10–50`, lease at `.30`, start changed to `.15`.  
   Expected: succeed. Same idea as shrinking the end, but on the start boundary.

4. **Expand start into free space on the left**  
   Range `20–50`, lease at `.30`, start changed to `.10` with free space below `20`.  
   Expected: succeed. Only the new strip on the left must be free.

5. **Shrink end so lease ends up outside pool**  
   Range `10–50`, lease at `.30`, end changed to `.25`.  
   Expected: succeed. The lease is no longer inside the range after the change.

All of the above passed.

**Expected to fail on resize:**

1. **Expand end over a lease outside the old pool**  
   Range `10–20`, lease at `.25`, end changed to `.30`.  
   Expected: rejected. The new strip `21–30` includes an address that is already in use.

2. **Expand start over a lease outside the old pool**  
   Range `20–50`, lease at `.15`, start changed to `.10`.  
   Expected: rejected. The new strip `10–19` includes an address that is already in use.

3. **Move pool to another subnet over a busy IP**  
   Range moved to another subnet where the new bounds overlap a busy address.  
   Expected: rejected. The resize exception should not apply; validation should run on the new subnet.

4. **Create new pool spanning an already-assigned IP**  
   New dynamic range `10–50` while `.30` is already assigned on the subnet.  
   Expected: rejected. The fix is for editing an existing range, not for creating one over a split gap.

All of the above were rejected as expected.

tanzin8r:
Nice work on the caching and subnet guard! Could you rebase onto upstream master when you get a chance? Some of the tests were failing are fixed there now.
On tests, I think you have the main bug covered already. The ones I would still add are the failure cases: expanding over a busy IP just outside the old pool on both start and end, a full clean() test when moving the range to another subnet that already has a clash, and one create-path test to make sure new ranges still get rejected if they span an already-assigned IP. 

On v3 side, the logic lives in src/maasapiserver/v3/api/public/models/requests/ipranges.py inside to_builder(), called from update_fabric_vlan_subnet_iprange() in the handler.  Subnet move is less of an issue there since the URL pins the subnet, but the resize-with-lease-inside-pool bug will still show up. We can create a seperate PR for that.

bryanfraschetti:
Thanks for the review!

Good idea to add tests for the failure cases. I just rebased and added some tests to cover the ones you mentioned: resizing, moving, or creating dynamic ranges over a busy IP.

I'll look into the v3 API and put up changes in a separate PR. Thank you for the pointers regarding that

tanzin8r:
lgtm. Thank you for the edits. Just one last test to make this concrete.

```
    def test_modify_existing_dynamic_range_can_expand_start_with_allocated_ip(
        self,
    ):
        subnet = make_plain_subnet()
        iprange = IPRange(
            subnet=subnet,
            type=IPRANGE_TYPE.DYNAMIC,
            start_ip="192.168.0.20",
            end_ip="192.168.0.50",
        )
        iprange.save()
        factory.make_StaticIPAddress(
            subnet=subnet,
            alloc_type=IPADDRESS_TYPE.AUTO,
            ip="192.168.0.30",
        )
        iprange.start_ip = "192.168.0.10"
        iprange.clean()
        iprange.save()
```

tanzin8r (approved):
+1

src/maasserver/models/iprange.py
```diff
@@ -274,18 +280,96 @@ def _validate_duplicates_and_overlaps(self):
...
+                return
+
+        if self._is_existing_dynamic_range_resize_allowed(
+            start_ip, end_ip, unused
+        ):
+            return
+
         self._raise_validation_error(message)
 
+    def _is_existing_dynamic_range_resize_allowed(
```
r00ta, line 298: why is this logic specific only for dynamic ranges? what about the other reserved ranges?

src/maasserver/models/iprange.py
```diff
@@ -274,18 +280,96 @@ def _validate_duplicates_and_overlaps(self):
...
+    ) -> bool:
+        """Allow resizing an existing dynamic range without re-querying the DB.
+
+        Uses cached original state (loaded_values) to compare the current
+        range against the persisted range, identifying only the newly added
+        segments (left/right expansion). Returns True if the range is merely
+        shrinking (no added segments) or if all newly added segments fit
+        within available unused ranges. Returns False if the original range
+        was not dynamic, the range moved to a different subnet, or any added
+        segment would overlap allocated IPs or other ranges.
```
r00ta, line 309: > would overlap allocated IPs

and where would this logic be in this function?

r00ta (changes requested):
questions inline

src/maasserver/models/iprange.py
bryanfraschetti: The other types of reserved ranges use different validation logic. This is specified in [L255](https://github.com/canonical/maas/blob/master/src/maasserver/models/iprange.py#L255) in src/maasserver/models/iprange.py

```
# Reserved ranges can overlap allocated IPs but not other ranges.
# Dynamic ranges cannot overlap anything (no ranges or IPs).
```

So Reserved ranges can overlap IPs and only hit a conflict when the range is overlapping another range. Because of that, resizing a reserved range is unaffected by allocations inside it. As a result when an IP address within a Reserved range was allocated, it would not affect the validation of the operation.

However, for dynamic range reservations, a resize should also be allowed in this case (because the IP allocation belongs to the existing reservation). However, the validation logic was incorrect and would reject such a resize because it viewed the allocated IP as splitting a continuous block of IPs and thought there was no way to modify the range reservation without overlapping an IP, which is not allowed.

src/maasserver/models/iprange.py
bryanfraschetti: In `_validate_duplicates_and_overlaps` the unused ranges are computed in this line:

```
            unused = self.subnet.get_ipranges_available_for_dynamic_range(
                exclude_ip_range_id=self.id
            )
```

That unused value is passed to this helper function and in this function we look at the proposed new range reservation boundaries and ensure that both are in unused ranges. Any extension to the left (eg. decreasing the lower bound of the range) must be in the same unused range as the current lower bound. Similar story for expansions to the right (increasing the upper bound of the reservation). If there is another allocated IP, the `get_ipranges_available_for_dynamic_range` will have split the available ranges on that IP, and then the following code block will return False (rejecting any proposed resize that would overlap an allocated IP) since the proposed boundary would not be in the same unused range as the current one

```
            if not any(
                added_start_ip in unused_range and added_end_ip in unused_range
                for unused_range in unused
            ):
                return False
```

r00ta (approved):
+1, thank you!

r00ta:
@bryanfraschetti there are test failures

bryanfraschetti:
Thanks @r00ta! I fixed the failing tests by rebasing

r00ta (approved):

skatsaounis:
@bryanfraschetti most probably you tried to fetch the latest master from GH UI. This created a commit with your personal email which has not signed the CLA. Most probably you use more than one emails in GH and the primary one is not Canonical affiliated or an email with which you have signed. Please take a look

bryanfraschetti:
@skatsaounis Thanks. I just signed the CLA with my personal email as that seemed easiest solution so it should pass now as the CLA doc says personal signatures are effective immediately

I'm not sure where it's picking up my personal email as the CI passes the CLA-check. I had rebased with: 
```
git fetch --all
git rebase origin/master
git push myfork --force
```
IIUC the action is coming from [here](https://github.com/canonical/has-signed-canonical-cla/blob/86d1373fc19075192a59dd06e8037abd5e44063d/index.js). It might be worth investigating why the differing behaviour between the two environments, but I'll look into that another time

PR #284, opened 2026-05-27
fix: check if in git repository when running clean-doc makefile target

alemar99:
The `clean-doc` makefile target in docs/ uses `git clean ...` but when building the deb package it fails since it's not in a git repository


maas-lander:
- [ ] Direct message
- [ ] ~maas

docs/Makefile
```diff
@@ -99,8 +99,11 @@ clean: clean-doc
 	rm -rf $(VALE_CONFIG)
 
 clean-doc:
-	git clean -fx "$(BUILDDIR)"
-	rm -rf $(SPHINXDIR)/.doctrees
+ifeq "$(shell git rev-parse --is-inside-work-tree 2>/dev/null | tr -d '\n' || true)" "true"
```
tmerten: Indentation looks odd are you sure that is correct?
Do you want to check for work-tree or `--is-inside-git-dir` (which is what the PR title suggests)?

tmerten (changes requested):
two quick questions, otherwise looking good.

docs/Makefile
alemar99: Yes that's odd, but that's how you use `ifeq` in makefiles, see [this example](https://www.gnu.org/software/make/manual/html_node/Conditional-Example.html)

alemar99: I've changed the git check to just run `git status` and rely on the status code

tmerten: :heart_eyes:

tmerten (approved):

PR #286, opened 2026-05-28
fix: properly wait maas-netmon in observe-arp

skatsaounis:
Replace `poll()` with `wait()` in maas-rack `observe-arp`.

`poll()` is non-blocking and returns `None` while the subprocess is still running, causing `run()` to return immediately before `maas-netmon` finishes. The Twisted parent then incorrectly treats the still-running child as having completed successfully. `wait()` blocks until the subprocess actually exits and always propagates the real exit code via `SystemExit`, ensuring the parent observes the true outcome.

This `poll()` call was effectively unreachable before commit https://github.com/canonical/maas/commit/25c89ad1c37c4ea75dc8d3c04b8bc06b73b188cf. Previously, `run()` blocked inside `observe_arp_packets()`, which read and parsed the monitor's output from a *separate* pipe for the entire lifetime of the process; the trailing `poll()` was only ever reached after that read loop had already finished. The switch to `maas-netmon` removed the `observe_arp_packets()` relay (the binary now writes JSON directly to the inherited stdout), leaving the non-blocking `poll()` as the only thing between spawning the child and returning — so the wrapper began exiting immediately while `maas-netmon` kept running. Restoring blocking behaviour via `wait()` makes the wrapper live for the full lifetime of `maas-netmon` again, as it did before that commit.

Resolves: LP:2151804

### DIFF
--- a/src/provisioningserver/utils/arp.py
+++ b/src/provisioningserver/utils/arp.py
@@ -51,6 +51,5 @@ def run(args, output=sys.stdout):
     cmd = sudo(cmd)
     network_monitor = subprocess.Popen(cmd, stdout=output)
     if network_monitor is not None:
-        return_code = network_monitor.poll()
-        if return_code is not None:
-            raise SystemExit(return_code)
+        return_code = network_monitor.wait()
+        raise SystemExit(return_code)
--- a/src/provisioningserver/utils/tests/test_arp.py
+++ b/src/provisioningserver/utils/tests/test_arp.py
@@ -31,11 +31,12 @@ class TestObserveARPCommand(MAASTestCase):
         add_arguments(parser)
         args = parser.parse_args(["eth0"])
         popen = self.patch(arp_module.subprocess, "Popen")
-        popen.return_value.poll = Mock()
-        popen.return_value.poll.return_value = None
+        popen.return_value.wait = Mock()
+        popen.return_value.wait.return_value = 0
         popen.return_value.stdout = io.StringIO("{}")
         output = io.StringIO()
-        run(args, output=output)
+        with self.assertRaisesRegex(SystemExit, "0"):
+            run(args, output=output)
         popen.assert_called_once_with(
             ["sudo", "-n", "/usr/sbin/maas-netmon", "eth0"], stdout=output
         )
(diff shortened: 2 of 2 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

skatsaounis:
# QA Steps

## Before the fix

Inspecting live processes:

```bash
root        5383  1.9  0.0      0     0 ?        Z    11:37   0:00 [python3] <defunct>
root        5386  0.0  0.0 1237608 12120 ?       Sl   11:37   0:00 /snap/maas/42013/usr/sbin/maas-netmon eth0
```

## After the fix

Inspecting live processes:

```bash
ubuntu@u1-bug:~$ ps aux | grep -E 'maas-netmon|observe-arp' | grep -v grep
root       10151  0.0  0.1  66600 55128 ?        S    11:42   0:00 python3 /snap/maas/x1/usr/bin/maas-rack observe-arp eth0
root       10154  0.0  0.0 1237608 12120 ?       Sl   11:42   0:00 /snap/maas/x1/usr/sbin/maas-netmon eth0
ubuntu@u1-bug:~$ sudo kill -9 10154
ubuntu@u1-bug:~$ ps aux | grep -E 'maas-netmon|observe-arp' | grep -v grep
root       11627  1.3  0.1  66600 55544 ?        S    11:56   0:00 python3 /snap/maas/x1/usr/bin/maas-rack observe-arp eth0
root       11628  0.0  0.0 1237352 11664 ?       Sl   11:56   0:00 /snap/maas/x1/usr/sbin/maas-netmon eth0
```

r00ta:
have you checked if there are other services that require the fix?

skatsaounis:
> have you checked if there are other services that require the fix?

This fix is specific to net-mon. It occurred with the introduction of https://github.com/canonical/maas/commit/25c89ad1c37c4ea75dc8d3c04b8bc06b73b188cf. I updated the PR description to explain further.

We have two other services `observe-beacons` and `observe-mdns` that are using the similar design with a separate PIPE. As long as we are not replacing them with maas-agent alternatives like netmon they are safe as is.

https://github.com/canonical/maas/blob/f172b5568ae223198ae6ed8d628b5b48734894e6/src/provisioningserver/utils/beaconing.py#L342

https://github.com/canonical/maas/blob/f172b5568ae223198ae6ed8d628b5b48734894e6/src/provisioningserver/utils/avahi.py#L259

r00ta (approved):

PR #290, opened 2026-05-29
docs: include _static dir

AloizioMacedo:


maas-lander:
- [ ] Direct message
- [ ] ~maas

PR #259, opened 2026-05-14
feat: migrate maas-ui from submodule to monorepo (master)

ndv99:
- Replace src/maasui/src submodule with git subtree from canonical/maas-ui main
- Update src/maasui/Makefile: SRC_DIR now points to BASE_DIR (content is inline), remove submodule update target, derive UI_REVISION from the MAAS repo commit, simplify build targets
- Add src/maasui/.gitignore entries for /tarballs and /nodejs; remove the three corresponding entries from root .gitignore
- Add ui paths-filter output to ci.yaml; update ci-check gate to include ui-lint and ui-test jobs
- Add UI CI workflows
- Update api-client-sync.yaml to use src/maasui in-tree instead of a separate canonical/maas-ui checkout; redirect generated PRs to canonical/maas
- Delete maas-ui-sync.yaml (daily submodule sync no longer needed)
- Added UI scopes to conventional commit checker

PR #292, opened 2026-05-29
feat: Add make targets to build snap from worktree

MaikRe:
Adds two new make targets and additional supporting functionality to make the UI and docs build works

`make snap-work-tree` functions essentially identical to `make snap-tree` expect that it cannot fetch the UI commit hash so it must be build locally.
`make snap-work-tree-sync` functions essentially identical to `make snap-tree-sync`.

Testing procedure:
 
 ```
 cd /path/to/your/maas/repo
 git worktree add ../new-worktree -b branch_name maik/snap-worktree # Specifically checkout this branch as obviously need the make targets
 cd ../new-worktree
 # To test from a truly clean environment
 make snap-clean
 make snap-tree-clean
 
 make snap-work-tree
 ```
 
 The work tree can then be mounted into an LXD container etc and updated with `make snap-work-tree-sync`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -470,6 +470,47 @@ snap:
 	$(snapcraft) pack
 .PHONY: snap
 
+snap-work-tree:
+	@if git rev-parse --is-inside-work-tree >/dev/null 2>&1; then \
+		git submodule update --init --recursive src/maasui/src; \
+	else \
+		echo "Skipping submodule update: git worktree metadata unavailable"; \
+	fi
+	$(MAKE) --no-print-directory snap-tree
+.PHONY: snap-work-tree
+
+snap-work-tree-sync:
+	@if git rev-parse --is-inside-work-tree >/dev/null 2>&1; then \
+		git submodule update --init --recursive src/maasui/src; \
+	else \
+		echo "Skipping submodule update: git worktree metadata unavailable"; \
+	fi
+	$(MAKE) --no-print-directory -C src/maasui build
+	$(MAKE) --no-print-directory go-bins
+	$(MAKE) --no-print-directory snap-tree
+	rsync -v -r -u -l -t -W -L --exclude 'maastesting' --exclude 'tests' --exclude 'testing' \
+		--exclude 'maasui' --exclude 'maasagent' --exclude 'maasopenfga' --exclude 'machine-resources' \
+		--exclude 'host-info' --exclude 'maas-offline-docs' \
+		--exclude '*.pyc' --exclude '__pycache__' \
+		src/ \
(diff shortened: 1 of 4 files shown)


maas-lander:
- [x] Direct message
- [ ] ~maas

docs/conf.py
```diff
@@ -16,10 +17,25 @@
 
 extensions = []
 
-if (
-    len(Path(__file__).parents) >= 2
-    and not (Path(__file__).parents[1] / ".git").exists()
-):
+
+def _has_usable_git_worktree() -> bool:
+    repo_root = Path(__file__).resolve().parents[1]
```
AloizioMacedo, line 22: You don't need to go to the repo root, `git rev-parse --is-inside-work-tree` should figure things out correctly (this would only be a potential problem in something like the maasui dir, which is a submodule).

So you can just use `cwd=Path(__file__).resolve().parents[0]`. Since we are already cleaning up the awkward check, this also gets a little more resilience since we don't need to assume we are inside another dir.  Theoretically there should be no problem in building the docs with the docs folder as root, in principle (e.g., if I mv everything of the docs folder inside the root of a docker container things could just work) and the current version would panic with a `IndexError` in that case. (Practically, we still use references to the package due to the API/CLI reference, but this will most likely change in the future, and there is no need to make extraneous assumptions where not necessary)

Makefile
```diff
@@ -470,6 +470,47 @@ snap:
...
+snap-work-tree:
+	@if git rev-parse --is-inside-work-tree >/dev/null 2>&1; then \
+		git submodule update --init --recursive src/maasui/src; \
+	else \
+		echo "Skipping submodule update: git worktree metadata unavailable"; \
+	fi
+	$(MAKE) --no-print-directory snap-tree
+.PHONY: snap-work-tree
+
+snap-work-tree-sync:
```
AloizioMacedo, line 482: With the changes proposed in place, those could be actually replacements of `snap-tree` and `snap-tree-sync`, right? They would work by having a standard `git clone`'d repo as well.

Maybe you are just trying to prevent overwriting the usual way due to potential unforeseen issues. In that case, maybe opening a new PR later with just the replacement itself? The point being that this is a kind of duplication.

MaikRe: > Maybe you are just trying to prevent overwriting the usual way due to potential unforeseen issues. In that case, maybe opening a new PR later with just the replacement itself? The point being that this is a kind of duplication.

Yes that is my primary issue. 
But for now I think I will table this until the monorepo #259 discussion is done as that might simplify some things as well.

PR #294, opened 2026-06-01
fix: retrieve the regions' ip addresses directly from the db in the image sync

alemar99:
The MAAS' system_users don't have any relationship in the openfga auth model and so, accessing public APIs result in unexpected results. Since in the other Temporal activities we are already using the db, fetch also this data from it.


maas-lander:
- [ ] Direct message
- [ ] ~maas

r00ta (approved):

PR #295, opened 2026-06-01
fix: block subnet delete when IPs are in use by nodes

tanzin8r:
backport for [LP#2138223](https://bugs.launchpad.net/maas/+bug/2138223)


maas-lander:
- [ ] Direct message
- [ ] ~maas

PR #296, opened 2026-06-02
docs: add google analytics

AloizioMacedo:
The Documentation team asked us to add google analytics support to our docs before merging it to prod.

Here is the page with the instructions: https://documentation.ubuntu.com/sphinx-stack/latest/how-to/optional-customisation/enable-google-analytics

Given that we have offline docs, I've done some conditional treatment to ignore those files for the offline build.


maas-lander:
- [ ] Direct message
- [ ] ~maas

AloizioMacedo:
Here is the request: https://code.launchpad.net/~aloiziomacedo/rtd-proxy-config/+git/rtd-proxy-config/+merge/505701/comments/1377433

docs/_templates/footer.html
```diff
@@ -0,0 +1,103 @@
+<div class="related-pages">
```
MaikRe, line 1: Don't know if we need the usual copyright header at the top of this page

docs/conf.py
```diff
@@ -7,6 +7,8 @@
 
 import yaml
 
+IS_ONLINE = not bool(os.getenv("MAAS_OFFLINE_DOCS", ""))
```
MaikRe: This just checks for this variable not being empty if I am not mistaken so even setting this to false in the snapcraft.yaml wouldn't change the resulting boolean.

MaikRe (commented):
A few inline comments/questions

docs/_templates/footer.html
AloizioMacedo: The file in the sphinx stack repo  is as such: https://github.com/canonical/sphinx-stack/blob/main/docs/_templates/footer.html, so let's keep as is.

MaikRe (approved):
+1

PR #297, opened 2026-06-02
chore: add django-piston3 as vendored dependency

AloizioMacedo:
Mainly intended to avoid having so much reliance on launchpad availability and speed.


maas-lander:
- [ ] Direct message
- [ ] ~maas

AloizioMacedo:
Adding as draft since it is unclear if we want to do this.

Thinking ahead, there is a good chance that our current way of building  the API reference and the CLI reference will make installing the MAAS dependencies for the docs unnecessary. (Imo, this is the ideal state.)

For instance, the API reference could be superseded when v3 is complete due to the existence of the Swagger /docs endpoint for FastAPI. 

The CLI reference could be done from the CLI source, which would not depend on a venv with MAAS due to the fact that it will be a separate Go project. It also might be simply unnecessary with proper CLI self-documentation.

r00ta (commented):
In general it's not a good approach to vendor this package and still rely on it in MAAS as a dependency. Long term it's a maintenance burden. We might want to use a different approach to build the docs, let's in the afternoon

AloizioMacedo:
After internal discussion, we decided to put django-piston3 in github and use that as dependency source instead.

PR #299, opened 2026-06-02
chore: reduce snap size with chisel slices and stage pruning

alexsander-souza:
- avoids staging the full deb dependency closure for packages that have chisel slices
- Prune artifacts from the maas part stage that serve no purpose in a snap.
- Audit and trim the Python stage-packages list
- Update required-packages/base to match

### DIFF
--- a/required-packages/base
+++ b/required-packages/base
@@ -49,6 +49,7 @@ python3-paramiko
 python3-passlib
 python3-petname
 python3-pexpect
+python3-prometheus-client
 python3-psycopg2
 python3-pydantic
 python3-pyinotify
@@ -65,9 +66,11 @@ python3-starlette
 python3-structlog
 python3-tempita
 python3-temporalio
+python3-twisted
 python3-txtftp
 python3-tz
 python3-ulid
+python3-uvicorn
 python3-uvloop
 python3-venv
 python3-yaml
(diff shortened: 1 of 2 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

r00ta:
what's the size of the snap before/after this patch?

nehjoshi (approved):
LGTM after doing some reading on how snap packaging works, though another review might be more useful.
I've verified that the 4 removed packages:
- `python3-protobuf`
- `python3-urllib3`
- `python3-pymacaroons`
- `python3-nacl`
are all transitive deps of other packages, and that the snap build succeeds without issues.

nehjoshi:
@r00ta 
> what's the size of the snap before/after this patch?

```sh
> ls -lh maas_*.snap
-rw-r--r-- 1 neh-joshi neh-joshi 263M Jun  3 11:42 maas_3.8.0~alpha1-18544-g.e2233a9d2_amd64_new.snap
-rw-r--r-- 1 neh-joshi neh-joshi 281M Jun  3 12:55 maas_3.8.0~alpha1-18545-g.2cde73299_amd64.snap
```
The "new" one refers to this branch

PR #301, opened 2026-06-02
fix: correct truncated pyright ignore in reserved IP request

tanzin8r:
(cherry picked from commit be620b8424c7838501f396465bb0353e381470a6)


maas-lander:
- [ ] Direct message
- [ ] ~maas

r00ta:
we have different lint rules across branches, you don't have to find a format that works for all the branches, just run make lint on your backport and push

r00ta (approved):

PR #302, opened 2026-06-02
chore: fix building of snap tree by using offline docs environment variable directly in make command

MaikRe:
Snapcraft doesn't like receiving environment variables in the snapcraft yaml for building the offline docs without analytics. This moves the environment variable directly into the make html call.


maas-lander:
- [x] Direct message
- [ ] ~maas

AloizioMacedo (approved):

PR #306, opened 2026-06-03
chore: pin aiohttp version to <3.14 to avoid change in init function

MaikRe:
See https://github.com/pnuckowski/aioresponses/issues/289 for details on what breaks in aiohttp 3.14.
aioresponses depends on it but does not pin the version to the one that does not break compatibility.


maas-lander:
- [x] Direct message
- [ ] ~maas

tanzin8r (dismissed):
+1

tanzin8r (approved):
+1

PR #307, opened 2026-06-03
fix: bump `starlette>=1.0.1` and use `scope["path"]`

lczyk:
addresses [CVE-2026-48710](https://nvd.nist.gov/vuln/detail/CVE-2026-48710) ( MAAS likely not vaunerable, but let's patch anyway )

`request.url.path` is derived from the Host header and can be forged by an attacker to bypass path-based middleware checks. scope["path"] comes from the ASGI request line and cannot be manipulated this way

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -111,7 +111,7 @@ sphinxcontrib-jsmath==1.0.1
 sphinxcontrib-qthelp==2.0.0
 sphinxcontrib-serializinghtml==2.0.0
 sphinxcontrib-svg2pdfconverter==2.1.0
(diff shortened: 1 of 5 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

r00ta:
thanks for the contribution. MAAS is not vulnerable because authentication is enforced at handler level. For the way `maasapiserver` is structured, the client can just send a request without any authentication header and hit the handler that will check for authentication. Our middleware is just adding the user to the context, so there is no point using the exploit to bypass the middleware, since they can already achieve the same by just omitting the headers.

PR #308, opened 2026-06-03
fix: remove deprecated squid config options

alexsander-souza:
deprecated since squid 6.x


maas-lander:
- [ ] Direct message
- [ ] ~maas

MaikRe (approved):
+1

PR #309, opened 2026-06-03
fix: bypass async_unsafe guard when closing DB connections in reactor thread

alexsander-souza:
Django's `connection.close()` raises `SynchronousOnlyOperation` when called
from a thread running an asyncio event loop, which happens in the Twisted
reactor thread. Call `_close()` directly since we are intentionally tearing
down connections and the guard is not needed here.

### DIFF
--- a/src/maasserver/utils/orm.py
+++ b/src/maasserver/utils/orm.py
@@ -37,6 +37,7 @@ from collections.abc import Iterable
 from contextlib import contextmanager, ExitStack
 from functools import wraps
 from itertools import chain, islice, repeat, takewhile
+import os
 import re
 import threading
 from time import sleep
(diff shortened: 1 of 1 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

r00ta (commented):
I'll look into this tomorrow

alexsander-souza:
this addresses this error during MAAS startup
```log
2026-06-02T17:49:44.456 12015052512 snap.maas.pebble.service maas-regiond[165086] INFO maasserver.eventloop_165086.worker: [info] Calling start_up to start region process
2026-06-02T17:49:44.457 12015053056 snap.maas.pebble.service maas-regiond[165086] INFO {"level": "critical", "message": "while handling timed call\n\tTraceback (most recent call last):\n\t  File \"/snap/maas/x1/usr/lib/python3/dist-packages/twisted/internet/base.py\", line 1105, in runUntilCurrent\n\t    call.func(*call.args, **call.kw)\n\t  File \"/snap/maas/x1/usr/lib/python3.14/dist-packages/maasserver/utils/orm.py\", line 919, in disable_all_database_connections\n\t    connection.close()\n\t  File \"/snap/maas/x1/usr/lib/python3/dist-packages/django/utils/asyncio.py\", line 24, in inner\n\t    raise SynchronousOnlyOperation(message)\n\tdjango.core.exceptions.SynchronousOnlyOperation: You cannot call this from an async context - use a thread or sync_to_async.\n\t", "name": "twisted.internet.base", "trace_id": "", "thread": "MainThread:137680342779712"}
```

r00ta:
I had a look. Let's talk on monday as I have a couple of points to discuss

alexsander-souza:
> I had a look. Let's talk on monday as I have a couple of points to discuss

I'm looking forward to it. I was hoping someone would have a better idea than what my limited knowledge of Twisted can come up with. (this fixes the issue but don't look like a proper solution)

src/maasserver/utils/orm.py
```diff
@@ -916,7 +916,13 @@ def disable_all_database_connections():
         connection = connections[alias]
         if type(connection) is not DisabledDatabaseConnection:
             connections[alias] = DisabledDatabaseConnection()
-            connection.close()
+            # connection.close() is guarded by @async_unsafe in Django, which
+            # raises SynchronousOnlyOperation when an asyncio event loop is
+            # running (e.g. the Twisted asyncio reactor thread). We bypass
+            # the guard by calling the underlying _close() directly, since
+            # we are intentionally tearing down connections here.
+            if connection.connection is not None:
```
r00ta: we call this method also outside the IO loop, maybe it's better to call `.close()` by default and fallback to `_close()` if we are inside the IO loop 

```
try:
    connection.close()
except SynchronousOnlyOperation:
    connection._close()
```

r00ta (approved):

PR #311, opened 2026-06-04
test: fix flaky test for node user data

alemar99:
The metadataserver fetches the user data for the user env when the node is deployed but the test only created user data for ephemeral env, resulting in a 404.


maas-lander:
- [ ] Direct message
- [ ] ~maas

r00ta (approved):

PR #312, opened 2026-06-04
refactor: return proper error code if script result does not exist (#237)

MaikRe:
If script_sets.get(id) returns None the error regarding an unknown id should only be return if that id does not match the naming of one of the three known ids. (current-commissioning, current-testing, current-installation), otherwise the function should provide clear information that no script set exists.


maas-lander:
- [x] Direct message
- [ ] ~maas

tanzin8r (approved):
+1

PR #315, opened 2026-06-05
chore: change source of django-piston3 to github

AloizioMacedo:
This should fix the issues we face due to availability problems with launchpad.


maas-lander:
- [ ] Direct message
- [ ] ~maas

alemar99 (approved):

PR #316, opened 2026-06-05
fix: preserve installation log pointer after machine release

tanzin8r:
Releasing a machine cleared current_installation_script_set, so the CLI
could not resolve current-installation even though install logs still
existed in the database.

Resolves [LP:2155055](https://bugs.launchpad.net/maas/+bug/2155055)


maas-lander:
- [x] Direct message
- [ ] ~maas

wyattrees (approved):
+1

PR #318, opened 2026-06-08
refactor: revert boot sources immutability

alemar99:
Let's postpone this change to MAAS 4.x, as it will break some automation.
Also, don't create the candidate boot source.


maas-lander:
- [ ] Direct message
- [ ] ~maas

r00ta (dismissed):
+1

r00ta (approved):

PR #319, opened 2026-06-08
fix: HTTP response body leak

troyanov:
The queueFlush function in src/maasagent/internal/dhcp/service.go never closes the HTTP response body returned by apiclient.Request()

Resolves: LP:2146799
(cherry picked from commit fcae4234e1a1f2c5506c1e5bc297996d45179ec7)


maas-lander:
- [ ] Direct message
- [ ] ~maas

PR #320, opened 2026-06-08
chore(ci): fix update-initial-sql workflow

alemar99:


maas-lander:
- [ ] Direct message
- [ ] ~maas

troyanov (approved):

PR #321, opened 2026-06-08
refactor: remove boot sources/resources dead code

alemar99:


maas-lander:
- [ ] Direct message
- [ ] ~maas

alexsander-souza (approved):

PR #322, opened 2026-06-08
chore: update starlette docs dep reference to 1.0.1

AloizioMacedo:
Helps in avoiding flagging from security scanners.

This creates a slight mismatch between what we actually use as dependency
on MAAS, and what we are using to just be able to generate the API/CLI
reference, but since there are no breaking changes (at least hitting
the code paths that the API/CLI reference walk through) this does not
create issues.


maas-lander:
- [ ] Direct message
- [ ] ~maas

alemar99 (approved):

PR #325, opened 2026-06-08
fix: preserve installation log pointer after machine release (#316)

tanzin8r:
Releasing a machine cleared current_installation_script_set, so the CLI could not resolve current-installation even though install logs still existed in the database.

Resolves [LP:2155055](https://bugs.launchpad.net/maas/+bug/2155055)

(cherry picked from commit 38fa551506d26403c8abbd58902c86c461d873d9)


maas-lander:
- [x] Direct message
- [ ] ~maas

alexsander-souza (approved):

PR #329, opened 2026-06-08
fix(dns): align snap BIND config with deb packaging

alexsander-souza:
bind9 9.20 (Ubuntu 26.04/Resolute) no longer ships the default db.* zone files that were previously installed in /etc/bind/.

Drop the explicit zones for 0.in-addr.arpa, 127.in-addr.arpa, and 255.in-addr.arpa from named.conf as they are covered by BIND's built-in empty zones (enabled by default since BIND 9.4).

Ship db.local directly in the snap and copy it alongside named.conf in run-named, since the localhost forward zone is not handled by BIND's built-in empty zones and has no package to provide it anymore.


maas-lander:
- [ ] Direct message
- [ ] ~maas

PR #330, opened 2026-06-08
fix(dns): align snap BIND config with deb packaging

alexsander-souza:
The snap was shipping static named.conf and db.local files that diverged from the system bind9 package and required manual maintenance. On Ubuntu 26.04+, bind9 no longer ships db.local and the localhost zone is handled internally by BIND.

Rewrite run-named to derive all BIND config files from the staged bind9 package at runtime, applying only the minimal snap-specific patches:
- named.conf: rewrite /etc/bind/ paths to $SNAP_DATA/bind/
- named.conf.options: written fresh with snap paths and the MAAS options include, replacing the trivial bind9-shipped version
- named.conf.root-hints: rewrite /usr/share/dns/ to $SNAP/usr/share/dns/
- named.conf.local: copied verbatim with the named.conf.maas include appended, mirroring what edit_named_options does for the deb install


maas-lander:
- [ ] Direct message
- [ ] ~maas

alexsander-souza:
there's no "browseable" version (that I know) of the config files shipped with `bind9`. the easiest way is to either:
1. install `bind9` in a resolute container 
2. download the [ubuntu patch](http://archive.ubuntu.com/ubuntu/pool/main/b/bind9/bind9_9.20.18-1ubuntu2.1.debian.tar.xz) and extract the files under `debian/extra/etc`

snap/local/tree/usr/bin/run-named
```diff
@@ -1,13 +1,34 @@
...
+sed "s|/etc/bind/|$SNAP_DATA/bind/|g" \
+    "$SNAP/etc/bind/named.conf" > "$SNAP_DATA/bind/named.conf"
+
+# Write named.conf.options with snap-specific paths.
+# This replaces the bind9-shipped file which only sets the cache directory.
+cat > "$SNAP_DATA/bind/named.conf.options" <<EOF
+options {
+    directory "$SNAP_DATA/bind/cache";
+    pid-file "$SNAP_DATA/bind/named.pid";
+    bindkeys-file "$SNAP/etc/bind/bind.keys";
```
AloizioMacedo, line 17: Might be worth removing this? It seems it [gets removed](https://github.com/isc-projects/bind9/blob/main/doc/changelog/changelog-9.21.15.rst) in a later version, and in the version we use it is [intended only for internal testing](https://github.com/isc-projects/bind9/blob/v9.20.18/lib/isccfg/namedconf.c#L1319).

Grepping for bind.keys in our repo, nothing shows up as well (other than this file).

AloizioMacedo (approved):

snap/local/tree/usr/bin/run-named
AloizioMacedo: By "this", I meant only the `bindkeys-file` line.

alexsander-souza: the goal is to keep the snap config on par with the deb, so I'd say we keep it until the deb removes it

PR #331, opened 2026-06-09
fix: do not retry transactions after losing the database connection

r00ta:
When `@transactional` retried after a retriable failure, `_reset()` reconnected to the database. But a new connectino does not hold the session level advisory lock acquired by `@synchronized` when it's used. This way, the retried code runs outside the critical section entirely unguarded.

On exit, `pg_advisory_unlock` fails on the new connection because it never held the lock, raising `DatabaseLockNotHeld` and crashing. In particular, this was a problem in the rack registration RPC handler. 

Remove the reconnection logic so that a broken connection simply fails the retry rather than silently running without the lock. 

Resolves LP:2156012


maas-lander:
- [ ] Direct message
- [ ] ~maas

alemar99:
Worth adding a test?

src/maasserver/utils/orm.py
```diff
@@ -747,25 +745,15 @@ def transactional(func):
     In addition, if `func` is being invoked from outside of a transaction,
     this will retry if it fails with a retryable failure.
     """
-
-    def _reset():
-        post_commit_hooks.reset()
```
alemar99, line 753: Re add this

alemar99 (approved):
LGTM, nice!

PR #332, opened 2026-06-09
fix: standard users should get can_view_available_machines entitlement when migrating to 3.8

r00ta:
`can_view_deployable_machines` was a leftover from one previous implementation. The right entitlement is `can_view_available_machines`


maas-lander:
- [ ] Direct message
- [ ] ~maas

nehjoshi (approved):
LGTM, thanks for the fix : )

troyanov (approved):

PR #333, opened 2026-06-09
fix(dhcp): remove dhcpd6-interfaces when no IPv6 interfaces are configured

alexsander-souza:
When no IPv6 interfaces are active, writing an empty dhcpd6-interfaces file still satisfies the systemd ConditionPathExists gate, causing dhcpd6 to attempt to start and fail. Remove the file instead so the service stays inactive.

Resolves LP:2156076

### DIFF
--- a/src/maasagent/internal/dhcp/service_test.go
+++ b/src/maasagent/internal/dhcp/service_test.go
@@ -430,6 +430,58 @@ func (s *DHCPServiceTestSuite) TestConfigureViaFile() {
 	}
 }
 
+// TestConfigureViaFileNoV4Interfaces ensures that when no IPv4 interfaces are
+// configured, the dhcpd-interfaces file is removed.
+func (s *DHCPServiceTestSuite) TestConfigureViaFileNoV4Interfaces() {
+	v4InterfacesPath := s.svc.dataPathFactory("dhcpd-interfaces")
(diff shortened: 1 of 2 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

src/maasagent/internal/dhcp/service_test.go
```diff
@@ -430,6 +430,32 @@ func (s *DHCPServiceTestSuite) TestConfigureViaFile() {
 	}
 }
 
+// TestConfigureViaFileNoV6Interfaces ensures that when no IPv6 interfaces are
+// configured, the dhcpd6-interfaces file is removed.
+func (s *DHCPServiceTestSuite) TestConfigureViaFileNoV6Interfaces() {
```
tanzin8r, line 461: Since we also remove dhcpd-interfaces when empty, consider adding test for v4 as well

tanzin8r (dismissed):
+1

tanzin8r (dismissed):
lgtm. The doc comment above T`estConfigureViaFileNoV4Interfaces`  describes the v6 test. if you prefer you can move it to T`estConfigureViaFileNoV6Interfaces` and add a matching comment for the v4 test. otherwise its fine

tanzin8r (approved):
+1

PR #334, opened 2026-06-09
docs: split docs django piston3 dependency

AloizioMacedo:
We need to use the launchpad dependency in order to build the deb. 

The split is not the prettiest thing, but it maintains things in `requirements.txt` files (which is what the [sphinx-stack](https://github.com/canonical/sphinx-stack/blob/main/docs/requirements.txt) uses), and allows us to avoid random build failures in the RTD builds.

We could theoretically do something a little cleaner with a `pyproject.toml`, but again this would deviate from the "standard" above. It is also a simple enough situation.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -36,4 +36,4 @@ formats:
 # Optionally declare the Python requirements required to build your docs
 python:
   install:
-  - requirements: docs/requirements.txt
+  - requirements: docs/requirements-github.txt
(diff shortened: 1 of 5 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

alemar99 (changes requested):
This would not solve the issue, we are still using pip and fetching things from the internet. The dependencies must be fetched from the ubuntu archive. We have to do what we are doing with MAAS and install dependencies through apt

AloizioMacedo:
@alemar99 I wonder if we can't simply build the docs outside of the deb environment and put the html bundle into it.

alemar99:
Uhm and how do you think we can achieve it? We can't fetch anything from the internet, the only thing that comes to my mind is having the built html folder in git but I don't think that's an option

AloizioMacedo:
@alemar99 By doing something similar to what happens with the UI and the go packages. I am continuing the work in that direction here: #360, please take a look when possible : )

PR #335, opened 2026-06-09
feat: track workflow operation status via local activity

tanzin8r:
Add a track_operation_status workflow decorator and supporting repository, service, and local activity layers to persist operation status transitions (RUNNING, COMPLETED, FAILED) in maasserver_operation. The decorator reads the OperationUUID search attribute and updates the operation record before and after the workflow run method executes.


maas-lander:
- [x] Direct message
- [ ] ~maas

src/maasservicelayer/db/repositories/operations.py
```diff
@@ -0,0 +1,55 @@
...
+    BaseExceptionDetail,
+    NotFoundException,
+)
+from maasservicelayer.exceptions.constants import (
+    UNEXISTING_RESOURCE_VIOLATION_TYPE,
+)
+from maasservicelayer.utils.date import utcnow
+
+
+class OperationsRepository(Repository):
```
alemar99: Why not a `BaseRepository`? In that case we would have all the basic methods already there

src/maasservicelayer/services/operations.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2026 Canonical Ltd.  This software is licensed under the
+# GNU Affero General Public License version 3 (see the file LICENSE).
+
+from maascommon.enums.operations import OperationStatus
+from maasservicelayer.context import Context
+from maasservicelayer.db.repositories.operations import OperationsRepository
+from maasservicelayer.services.base import Service
+
+
+class OperationsService(Service):
```
alemar99: Same as above, why not a `BaseService`?

alemar99 (changes requested):
See inline

src/maasservicelayer/db/repositories/operations.py
tanzin8r: Per the spike, operations were identified by uuid. I figured it would be best to use this instead of `Baserepository`. But I'll update it to `BaseRepository` since that was a discrepancy in the spike.

src/maastemporalworker/workflow/__init__.py
```diff
@@ -16,4 +16,7 @@
 When you create a new workflow or a new activity, you MUST NOT use @workflow.run nor @activity.defn. Instead, you should use
 the decorators @workflow_run_with_context and @activity_defn_with_context provided in maastemporalworker.workflow.utils. These
 decorators are wrapping the temporal decorators with the contextual logger.
+
+For workflows that track a long-running operation in the database, use @workflow_run_with_tracked_operation instead of
+@workflow_run_with_context. It combines context setup and operation status tracking into a single decorator.
```
MaikRe: Is it not possible to apply two decorators to a workflow?/ Is there every the case where we might want to have the tracker operation workflow but not the context?

tanzin8r: that was my earlier approach. I had two separate decorators stacked on the workflow. I moved to a single combined decorator mainly as a design decision after talking with Wyatt. Basically `workflow_run_with_context` is required for every workflow because it registers the run method with Temporal and sets up trace/logging. track_operation_status only handles updating the operation in the database, so there isn’t really a production case where we’d want tracking without context.

src/tests/maastemporalworker/workflow/test_utils.py
```diff
@@ -1,11 +1,92 @@
-# Copyright 2025 Canonical Ltd.  This software is licensed under the
+# Copyright 2026 Canonical Ltd.  This software is licensed under the
```
MaikRe:
```suggestion
# Copyright 2025-2026 Canonical Ltd.  This software is licensed under the
```

src/maascommon/workflows/operation.py
```diff
@@ -0,0 +1,18 @@
...
+# GNU Affero General Public License version 3 (see the file LICENSE).
+
+import dataclasses
+
+from maascommon.enums.operations import OperationStatus
+
+# Search attribute used to link a workflow execution to its operation.
+OPERATION_UUID_SEARCH_ATTRIBUTE = "OperationUUID"
+
+UPDATE_OPERATION_STATUS_ACTIVITY_NAME = "update-operation-status"
```
MaikRe: These constants should not be define in maascommon but in the module itself. Have a look at how other activities do this.

src/maasservicelayer/services/operations.py
```diff
@@ -0,0 +1,49 @@
...
+    BaseService[Operation, OperationsRepository, OperationBuilder]
+):
+    def __init__(
+        self,
+        context: Context,
+        operations_repository: OperationsRepository,
+    ):
+        super().__init__(context, operations_repository)
+
+    async def update_status(
```
MaikRe, line 27: This should either overwrite BaseService.update_one, call it here, or use it directly with proper update of the resource semantics. By writing a custom method and then calling the repository update one directly you are bypassing the pre and post update hooks which doesn't follow the convention of the repo

src/tests/maastemporalworker/workflow/test_operation.py
```diff
@@ -0,0 +1,56 @@
...
+import maastemporalworker.workflow.activity as activity_module
+from maastemporalworker.workflow.operation import OperationActivity
+
+ERROR_MESSAGE = "operation failed"
+
+
+@pytest.mark.asyncio
+@pytest.mark.usefixtures("maasdb")
+class TestOperationActivity:
+    async def test_update_operation_status(
```
MaikRe, line 31: You only test the successful path here, what happens if the operation_uuid id isn't found?
I know that shouldn't be the case but since we are function based on string matching it might be nice to handle this

MaikRe (changes requested):

src/maascommon/workflows/operation.py
tanzin8r: I moved activity constants into the activity module but the decorator that calls that activity still lives alongside the other generic workflow decorators. The activity module imports the generic decorator helper from the utils module and the decorator needs those constants. I am using lazy import to resolve circular import.

src/maastemporalworker/workflow/utils.py
```diff
@@ -1,20 +1,23 @@
...
 from temporalio import activity, workflow
 from temporalio.exceptions import ApplicationError, TemporalError
 
+from maascommon.enums.operations import OperationStatus
 from maascommon.tracing import get_or_set_trace_id
 from maasservicelayer.context import Context
 
 logger = structlog.getLogger()
 
+UPDATE_OPERATION_STATUS_TIMEOUT = timedelta(seconds=30)
```
MaikRe: Other timeouts are defined in their respective module. Is there a reason this one is defined in utils.py?
Consider for example "POWER_ACTION_ACTIVITY_TIMEOUT"

src/maastemporalworker/workflow/utils.py
```diff
@@ -63,9 +66,81 @@ async def wrapper(*args, **kwargs):
...
+    """
+    You MUST use this decorator instead of @workflow_run_with_context
+    on operation-tracked workflows.
+    """
+    return workflow_run_with_context(track_operation_status(func))
+
+
 def async_retry(retries=5, backoff_ms=1000):
     def wrapper(fn):
-        @functools.wraps(fn)
```
MaikRe, line 68: How is this change related to the wrapper?

src/maastemporalworker/workflow/utils.py
```diff
@@ -63,9 +66,81 @@ async def wrapper(*args, **kwargs):
...
+            await workflow.execute_local_activity(
+                UPDATE_OPERATION_STATUS_ACTIVITY_NAME,
+                UpdateOperationStatusParam(
+                    operation_uuid=operation_uuid,
+                    status=OperationStatus.FAILED,
+                    error=str(e),
+                ),
+                start_to_close_timeout=UPDATE_OPERATION_STATUS_TIMEOUT,
+            )
+            raise
```
MaikRe: I don't know enough about this but the logic here worries me a little. If temporal throws an exception are we sure the local activity to update the operation status will still execute deterministically and correctly? Maybe @alemar99 can confirm this is the correct way to do this

MaikRe (changes requested):
A few more questions/comments

src/maastemporalworker/workflow/utils.py
tanzin8r: It's not. I only changed it because wraps was already imported for the other decorators, and `async_retry` was using functools.wraps directly. Switched it over so I could drop the extra `import functools`

src/maastemporalworker/workflow/utils.py
tanzin8r: Fair point. I put it in `utils.py` because the code that calls `execute_local_activity` is `track_operation_status` in that file, not in module. Power have workflow and activity together. here the decorator is generic and lives with the other workflow helpers. I am thinking if it would be even better to move workflow into the module as well. That way I can get rid of lazy import as well.

src/maastemporalworker/workflow/utils.py
tanzin8r: I had the same concern. I am following the same approach as POC. On failure you catch the exception, call a local activity to mark FAILED, then re-raise. That `execute_local_activity` call gets recorded in workflow history, so on replay Temporal replays it from history instead of running it again. I added a test for this as well
One gap is external termination, if that happens the except block won't run then we'll need a reconciliation workflow. I left that out on purpose from this PR.

alemar99: As I remember [from the poc](https://github.com/alemar99/temporal-long-running-ops-status-tracking) I did some time ago, this should be the right way of doing it and what @tanzin8r is saying should be correct. I'd advise to make sure this is the case before merging it.

src/maastemporalworker/workflow/utils.py
MaikRe: Nitpick but could you do a separate PR for this? Its a small change but for the sake of code cleanliness I would like to avoid scope creep

MaikRe (dismissed):
+1

src/maastemporalworker/workflow/operation.py
```diff
@@ -0,0 +1,114 @@
...
+            await workflow.execute_local_activity(
+                UPDATE_OPERATION_STATUS_ACTIVITY_NAME,
+                UpdateOperationStatusParam(
+                    operation_uuid=operation_uuid,
+                    status=OperationStatus.RUNNING,
+                ),
+                start_to_close_timeout=UPDATE_OPERATION_STATUS_TIMEOUT,
+            )
+
+            result = await func(self, param)
```
alemar99, line 80: I think that result should be saved in the operation in the db? (Through the `update_status` activity)

src/maasservicelayer/services/operations.py
```diff
@@ -0,0 +1,49 @@
...
+        builder = OperationBuilder(status=status)
+        if status == OperationStatus.RUNNING:
+            builder.started = utcnow()
+        elif status in (
+            OperationStatus.COMPLETED,
+            OperationStatus.FAILED,
+        ):
+            builder.finished = utcnow()
+        if error is not None:
+            builder.result_errors = {"error": error}
```
alemar99: add `result` as well

src/maastemporalworker/workflow/operation.py
```diff
@@ -0,0 +1,114 @@
...
+# Activities names
+UPDATE_OPERATION_STATUS_ACTIVITY_NAME = "update-operation-status"
+
+
+# Activities parameters
+@dataclass
+class UpdateOperationStatusParam:
+    operation_uuid: str
+    status: OperationStatus
+    error: str | None = None
```
alemar99, line 35: add result

alemar99 (commented):
You're not saving the result of the operation in the db

alemar99 (approved):
+1

src/maastemporalworker/workflow/operation.py
```diff
@@ -0,0 +1,117 @@
...
+        async with self.start_transaction() as services:
+            await services.operations.update_status(
+                operation_uuid=param.operation_uuid,
+                status=param.status,
+                result=param.result,
+                error=param.error,
+            )
+
+
+def track_operation_status(func):
```
r00ta, line 52: why would you blindly track every single activity of the workflow? Shouldn't the workflow itself decide how and when the operation should be updated? 

Using a decorator to do this will lead to designing the workflow in a 1:1 mapping between the actual activities and the status of the operations

r00ta: never mind, I misunderstood the code at the first look

tanzin8r: The decorator is wrapping the workflow `run()` method one time and only writes the operation status at the boundaries of the run. Whatever activities the workflow executes in between are invisible to it. It cannot create an activity-to-status mapping, since it never sees the activities. The operation status reflects the lifecycle of the operation as a whole, not the steps inside it.

According to the spec, for start, finish and error it proposes instrumenting the run function via a decorator with a local activity. For the current task, the status must be updated manually before starting each task, with the workflow executing an activity at the points it chooses. I have a seperate ticket for that and that's the next step. Are you asking if transition states should be written explicitly inside each workflow?

PR #336, opened 2026-06-10
fix: display the right message when logging in LP#2155376

nehjoshi:
See the following table for the possible scenarios that have been implemented:

| Case | OIDC enabled? | Account state                                  | Response                                              |
|------|---------------|------------------------------------------------|-------------------------------------------------------|
| A1‑I | yes           | profile exists, **bound** to a provider        | `200 { is_oidc: true }` + auth URL + state/nonce cookies |
| A1‑II| yes           | profile exists, **not** bound to a provider    | `200 { is_oidc: false }` (login via password auth) |
| A1‑III| yes           | profile exists, bound to a **different** provider than the enabled one | `409 Conflict` (account is reserved for a provider that is not currently enabled) |
| A2   | yes           | username/email does **not** exist                  | `200 { is_oidc: true }` + auth URL + state/nonce cookies (new users are provisioned at OIDC callback) |
| B1‑I | no            | profile exists **and** bound to a provider     | `409 Conflict` (account is reserved for OIDC and OIDC is off) |
| B1‑II| no            | profile exists, **not** bound to a provider    | `200 { is_oidc: false }` (login via password auth)                             |
| B2   | no            | username/email does **not** exist                  | `200 { is_oidc: false }` (login via password auth, which will eventually result in 401 after the user enters password, because the profile doesn't exist) |

Resolves: LP:2155376

Resolves [MAASENG-6544](https://warthogs.atlassian.net/browse/MAASENG-6554)

[MAASENG-6544]: https://warthogs.atlassian.net/browse/MAASENG-6544?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


maas-lander:
- [x] Direct message
- [ ] ~maas

alemar99:
Question: what happens if the user is connected to a different provider which is not the currently enabled one? Is this a possible scenario?

nehjoshi:
> Question: what happens if the user is connected to a different provider which is not the currently enabled one? Is this a possible scenario?

Yes, it is, good catch! I will add this edge case in, as well...

nehjoshi:
Added the case for when a user profile is associated with a different provider than the one currently enabled (see table in the PR description which has been updated)

src/maasapiserver/v3/api/public/handlers/auth.py
```diff
@@ -177,13 +178,48 @@ async def initiate_auth_flow(
...
+                details=[
+                    BaseExceptionDetail(
+                        type=MISSING_PROVIDER_CONFIG_VIOLATION_TYPE,
+                        message=(
+                            "This account is linked to an OIDC provider "
+                            "that is currently disabled."
+                        ),
+                    )
+                ]
             )
```
alemar99, line 205: Maybe easier to read approach?

```python
# local user
if user_profile is not None and user_profile.provider_id is None:
    return AuthInfoResponse(is_oidc=False)
# OIDC user but either no provider or wrong provider
if user_profile is not None and user_profile.provider_id is not None:
    if provider is None or provider.id != user_profile.provider_id:
        raise ConflictException(
            details=[
                BaseExceptionDetail(
                    type=MISSING_PROVIDER_CONFIG_VIOLATION_TYPE,
                    message=(
                        "This account is linked to an OIDC provider "
                        "that is currently disabled."
                    ),
                )
            ]
        )
# no provider, only local login
if provider is None:
    return AuthInfoResponse(is_oidc=False)

# provider enabled
```

alemar99:
(I've edited your commit description to change the bug reference to [what we expect in the GH action](https://github.com/canonical/maas/blob/master/.github/workflows/lp-bug-update.yaml#L24))

nehjoshi:
The first `if(...)` block takes care of cases A1-III and B1-I.
The second `if(...)` block takes care of cases A1-II, B1-II, and B2.
The code after that takes of cases A1-I and A2.

alemar99 (approved):
Nice, LGTM

PR #337, opened 2026-06-10
fix(power): don't run power activities for power drivers of type manual

alemar99:
Resolves: LP:2153152

### DIFF
--- a/src/maasserver/models/node.py
+++ b/src/maasserver/models/node.py
@@ -6180,7 +6180,10 @@ class Node(CleanSave, TimestampedModel):
             self._start_deployment()
             claimed_ips = True
             needs_power_call = False
-            task_queue = str(get_temporal_task_queue_for_bmc(self))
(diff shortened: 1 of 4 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

src/maasserver/models/node.py
```diff
@@ -6180,7 +6180,10 @@ def claim_auto_ips(_):
             self._start_deployment()
             claimed_ips = True
             needs_power_call = False
-            task_queue = str(get_temporal_task_queue_for_bmc(self))
+            if power_info.power_type == "manual":
+                task_queue = ""
```
r00ta: I would add a small comment to explain this, even if it might be obvious for most of the developers/agents

r00ta (commented):
looks good, one nitpit

r00ta (approved):

PR #338, opened 2026-06-10
refactor: clean up operations tables to match naming conventions

MaikRe:


maas-lander:
- [x] Direct message
- [ ] ~maas

alemar99 (approved):
+1

PR #339, opened 2026-06-10
feat: operations service layer, repository layer, and handler for GET operations

MaikRe:


maas-lander:
- [x] Direct message
- [ ] ~maas

src/maasapiserver/v3/api/public/handlers/operations.py
```diff
@@ -0,0 +1,132 @@
...
+            200: {
+                "model": OperationsListResponse,
+            },
+        },
+        response_model_exclude_none=True,
+        status_code=200,
+        dependencies=[
+            Depends(
+                check_permissions(
+                    MAASResourceEntitlement.CAN_VIEW_GLOBAL_ENTITIES
```
r00ta: so if you can view global entities you can also see the deployments of other users?

r00ta (commented):
Questions around authorization inline

src/maasapiserver/v3/api/public/handlers/operations.py
MaikRe: Ah good point, the spec defined this in admin/nonadmin terms which I forgot doesn't directly translate in the permissions. @alemar99 or @motrilense what sort of permissions were you intending when the spec was written? Should this be a new entitlement for "admins" to see all users operations?

MaikRe: I would suggest to add CAN_VIEW_OPERATIONS and in the future also CAN_EDIT_OPERATIONS (for the DELETE endpoint) and handle the filtering of userId vs seeing all operations the same way this is currently done for CAN_VIEW_DNSRECORDS

r00ta: How can the user monitor his own operations then?

alemar99: Adding a `CAN_VIEW_OPERATIONS` for admins might be a solution, but not in the `check_permissions` function, the check must be done in the handler. If a user has the entitlement `CAN_VIEW_OPERATIONS` we return the full list of operations, otherwise we filter operations based on the `user_id` db column of the OperationsTable. (IIUC this is the same as Maik was suggesting)

MaikRe: > How can the user monitor his own operations then?

Yes @alemar99 summarized exactly what I was suggesting! I will implement that then

src/maasservicelayer/services/operations.py
```diff
@@ -50,3 +60,56 @@ async def update_status(
...
+                    BaseExceptionDetail(
+                        type=UNEXISTING_RESOURCE_VIOLATION_TYPE,
+                        message=f"Operation with uuid '{uuid}' was not found.",
+                    )
+                ]
+            )
+        if can_view_all:
+            return operation
+        if operation.user_id != user_id:
+            raise UnauthorizedException(
```
alemar99: Just raise a `NotFoundException`.

src/tests/maasservicelayer/services/test_operations.py
```diff
@@ -50,6 +71,145 @@ def test_instance(self) -> MaasBaseModel:
...
+        query = QuerySpec()
+        obj = await service_instance.update_one(
+            query, builder, test_instance.etag()
+        )
+        assert obj == test_instance
+        service_instance.repository.update_by_id.assert_awaited_once_with(
+            id=test_instance.id, builder=builder
+        )
+
+    async def test_update_one_etag_not_matching(
```
alemar99, line 109: Why these are not skipped?

src/maasapiserver/v3/api/public/models/requests/operations.py
```diff
@@ -0,0 +1,52 @@
...
+
+from fastapi import Query
+from pydantic import BaseModel, Field
+
+from maascommon.enums.operations import OperationStatus, OperationType
+from maasservicelayer.db.filters import Clause
+from maasservicelayer.db.repositories.operations import OperationsClauseFactory
+
+
+class OperationFilterParams(BaseModel):
```
alemar99, line 12: Missing tests

src/maasapiserver/v3/api/public/models/responses/operations.py
```diff
@@ -0,0 +1,68 @@
...
+    BaseHal,
+    BaseHref,
+    HalResponse,
+    PaginatedResponse,
+)
+from maascommon.enums.operations import OperationStatus, OperationType
+from maasservicelayer.models.operations import Operation
+
+
+class OperationResponse(HalResponse[BaseHal]):
```
alemar99, line 19: Missing test

alemar99 (changes requested):
See inline

src/tests/maasservicelayer/services/test_operations.py
MaikRe: That was an oversight, I had implemented them as sort of boilerplate code but meant to remove/skip them as they are not part of this PR. They have no been removed.

alemar99 (approved):
+1

alemar99:
I've deleted the description otherwise it would be included in the commit and it was not relevant

PR #341, opened 2026-06-10
feat(security): add FIPS mode detection and replace non-FIPS crypto primitives

alexsander-souza:
Introduce maascommon.fips module with host FIPS mode detection via /proc/sys/crypto/fips_enabled, FIPS-approved SSH algorithm configs, and SSH key validation helpers.

Replace Fernet (AES-CBC + HMAC-SHA256 construction) with AES-256-GCM for pre-shared key encryption in provisioningserver.security, as AES-GCM is directly FIPS 140-3 approved. Update all consumers (beaconing, RPC region/cluster services).

Add FIPS policy enforcement for certificate key generation (reject DSA, enforce minimum RSA 2048-bit keys).

Replace PostgreSQL md5() in BMC unique index with sha256() via Alembic migration, since md5() is blocked under FIPS mode.

Mark non-security SHA-1 usages with usedforsecurity=False (API doc hashing, WebSocket accept, test factory) and remove MD5 password hasher from Django development settings when FIPS is active.

Add FIPS structured-log event constants for consistent observability.


maas-lander:
- [ ] Direct message
- [ ] ~maas

src/maasserver/djangosettings/development.py
```diff
@@ -109,10 +110,13 @@
 else:
     MIGRATION_MODULES["maastesting"] = "maastesting.migrations"
 
-PASSWORD_HASHERS = (
+_base_hashers = (
     "django.contrib.auth.hashers.MD5PasswordHasher",
     "django.contrib.auth.hashers.PBKDF2PasswordHasher",
 )
+PASSWORD_HASHERS = tuple(
+    h for h in _base_hashers if not (is_fips_enabled() and "MD5" in h)
```
r00ta, line 118: I assume we don't support moving an environment initialized without fips to an env with fips enabled. Correct?

src/maasservicelayer/db/alembic/versions/0023_replace_bmc_md5_index_with_sha256.py
```diff
@@ -0,0 +1,51 @@
...
+        ON maasserver_bmc USING btree (
+            power_type,
+            sha256(power_parameters::text::bytea)
+        )
+        WHERE ((power_type)::text <> 'manual'::text);
+        """
+    )
+
+
+def downgrade() -> None:
```
r00ta, line 40: no need, we don't support downgrading

src/provisioningserver/certificates.py
```diff
@@ -29,11 +31,38 @@
...
+        logger.error(
+            FIPS_CRYPTO_ERROR,
+            operation="key_generation",
+            algorithm=f"RSA-{key_bits}",
+            reason=("RSA key size below FIPS minimum of 2048 bits"),
+        )
+        raise CertificateError(
+            f"RSA key size {key_bits} is below FIPS minimum of 2048 bits"
+        )
+
```
r00ta, line 64: Do we need to validate other types or do we let them through without checks?

src/provisioningserver/security.py
```diff
@@ -89,69 +87,99 @@ def _get_or_create_fernet_psk() -> bytes:
                 backend=default_backend(),
             )
             key = kdf.derive(secret)
-            key = urlsafe_b64encode(key)
```
r00ta, line 92: why `key = urlsafe_b64encode(key)` has been removed?

src/provisioningserver/security.py
```diff
@@ -89,69 +87,99 @@ def _get_or_create_fernet_psk() -> bytes:
...
+    all base64-encoded. The timestamp is bound to the ciphertext as AAD.
     """
-    fernet = _get_fernet_context()
+    aesgcm = _get_aesgcm_context()
     if isinstance(message, str):
         message = message.encode("utf-8")
-    token = fernet.encrypt(message)
+    nonce = os.urandom(12)
+    ts = int(time.time()).to_bytes(8, "big")
+    ciphertext = aesgcm.encrypt(nonce, message, ts)
```
r00ta, line 114: do you have performance tests around the new encryption vs the fernet one?

r00ta (commented):
some questions inline

PR #342, opened 2026-06-11
fix: reconfigure-pebble is under usr/bin/

r00ta:
In 3.8 the `reconfigure-pebble` binary is under `usr/bin`, not `bin/`.


maas-lander:
- [ ] Direct message
- [ ] ~maas

troyanov (approved):

PR #344, opened 2026-06-11
fix: revert validate ip address in machine power config

troyanov:
This reverts commit 4950ce99bdbc94885063abfc747f4c8717e3ec04.

Resolves: LP:2150892


maas-lander:
- [ ] Direct message
- [ ] ~maas

PR #346, opened 2026-06-11
fix: bump host-info LXD dependency to 6.8 to fix duplicate disk serial numbers

r00ta:
LXD 5.21 returns the same SCSI serial for all disks behind the same controller on certain HP machines, causing curtin to crash during deployment as it cannot distinguish devices (LP:2102135)
It also fails to expose proper disk identifiers in virtual environments, casing MAAS to return unstable paths like /dev/vdb instead of stable /dev/disk/by-id/ paths (LP:2087945)

LXD 6.8 correctly reports unique per-disk serials and proper device identifiers, resolving both issues.

This also enables recommissioning with either Jammy or Noble for Resolute deployments, because the old LXD fork running on resolute returns the same serials of the new LXD 6.8 running on all the ubuntu releases.

Resolves LP:2087945
Resolves LP:2102135


maas-lander:
- [ ] Direct message
- [ ] ~maas

alexsander-souza (approved):

PR #348, opened 2026-06-11
feat: add operations permissions to openfga

MaikRe:
Adds permissions for viewing and editing all long running operations as defined by the 'admin' permission in MA276. Users without this permission will only be able to view and cancel operations which contain their user ID.


r00ta:
say a user starts the deployment of a machine, how do you check that the user can actually view their own operations?

maas-lander:
- [x] Direct message
- [ ] ~maas

MaikRe:
> say a user starts the deployment of a machine, how do you check that the user can actually view their own operations?

Similar to how we currently do the check for CAN_VIEW_DNSRECORDS the API handler checks whether the user has the above named permissions. If the user does they can view all operations regardless of owner. We do this by passing a parameter to the service layer which depending the case either filters the operations returned by user ID or shows all of them.

In the case of GET /operations this may return an empty list if the user has no operations and does not have the CAN_VIEW_OPERATIONS permission

In the case of GET /operations/{uuid} should the user provide an operation UUID of which they are not the owner they will receive a forbidden error (403)

docs/how-to-guides/user-groups-and-entitlements.md
```diff
@@ -81,6 +81,8 @@ Permissions follow two inheritance rules:
 | `can_view_devices` | View all devices. |
 | `can_view_ipaddresses` | View all IP addresses. |
 | `can_view_dnsrecords` | View all DNS resource record sets. |
+| `can_view_operations` | View all Operations. |
+| `can_edit_operations` | Edit all operations. |
```
r00ta: there is no change this is feature is going to land into 3.8. This should not be exposed to users

src/maasopenfga/internal/migrations/00003_add_can_view_operations.go
```diff
@@ -0,0 +1,182 @@
...
+
+type user
+
+type group
+  relations
+    define member: [user]
+
+type maas
+  relations
+    define can_edit_machines: [group#member]
```
r00ta: no need for this. Just edit the first two migrations accordingly. Nobody should use latest/edge so it's fine to just ask people to recreate the DB.

r00ta (changes requested):

src/maasopenfga/internal/migrations/00003_add_can_view_operations.go
MaikRe: Hasn't the cut for 3.8 already started through, at least for the beta? Editing the original migration files seems like this would break whats already in progress for that?

MaikRe: Also the connection to maasserver_usergroup_members_view makes the manual surgery needed for existing maas-test-db instances more difficult as the process would need to revert alembic migrations as well.

Happy to make this change, just not sure if its the cleanest way to do this.

r00ta: > Hasn't the cut for 3.8 already started through

No, 3.8 hasn't been cut yet

> Editing the original migration files seems like this would break whats already in progress for that?

Nope

r00ta: actually, I'd argue that the operations work should not have landed partially in 3.8 at all. Say you'll discover something on the way and you need to change older stuff, then you'll probably end up dropping the tables and recreating them. But anyways it's not a big deal, just a bit more work to do in the future

MaikRe:
Adding this as a note for how to properly migrate dev environment:


```
# Accessing the installed test db
sudo maas-test-db.psql

DROP TABLE IF EXISTS public.maasserver_machine_operation;
DROP TABLE IF EXISTS public.maasserver_operation_task;
DROP TABLE IF EXISTS public.maasserver_operation;

ALTER TABLE IF EXISTS public.maasserver_interface DROP COLUMN IF EXISTS switch_id;
DROP TABLE IF EXISTS public.maasserver_switch;
DROP SCHEMA IF EXISTS openfga CASCADE;
UPDATE public.alembic_version SET version_num = '0019';
\q

sudo maas migrate

```

r00ta (approved):
lgtm

PR #350, opened 2026-06-11
fix: handle exceptions when decrypting cookies and log a warning

alemar99:
CookieManager.decrypt method can raise exceptions in the following scenarios:
- if the cookie is not of the expected length, a `ValueError` is raised
- if the cookie is not cryptographically valid, an `InvalidTag` is raised

In any case, log a warning since the cookie was probably manually altered.

### DIFF
--- a/src/maasapiserver/v3/auth/cookie_manager.py
+++ b/src/maasapiserver/v3/auth/cookie_manager.py
@@ -3,10 +3,15 @@
 
 from enum import StrEnum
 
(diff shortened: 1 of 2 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

nehjoshi (dismissed):
+1 Nice finding!

nehjoshi (approved):
LGTM

PR #353, opened 2026-06-12
docs: allow downloading the html bundle from RTD

AloizioMacedo:
This makes it easier for someone to have an offline version if they want, independently from having our own "bundled" version inside the snap/deb.


maas-lander:
- [ ] Direct message
- [ ] ~maas

alemar99 (approved):
+1

PR #357, opened 2026-06-12
feat(dhcp): generate Kea shared network config

wyattrees:


maas-lander:
- [ ] Direct message
- [ ] ~maas

r00ta (changes requested):
Not sure why this MP is open. We can't switch to kea at the moment and we were supposed to do a spike in order to write a spec. I'd not focus on implementing everything at this stage

r00ta (commented):
My bad, I did not notice this was not targeting master. Ignore my comment, all good

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -654,7 +689,386 @@ async def _get_dhcp_host_reservations(
...
+        svc: ServiceCollectionV3,
+    ) -> list[str]:
+        if not subnet.allow_dns:
+            return []
+
+        def get_IPvAnyAddress_from_IPAddress(ip: IPAddress) -> IPvAnyAddress:
+            return (
+                IPv4Address(str(ip))
+                if ip.version == 4
+                else IPv6Address(str(ip))
```
AloizioMacedo:
```suggestion
                IPv4Address(ip.value)
                if ip.version == 4
                else IPv6Address(ip.value)
```

`ip.value` should be just an int, so this should have less overhead, even if here it doesn't have much material impact. (Funnily enough, in the past there were issues with performance of this kind of conversion on a hot loop, so it is good to set examples.)

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -654,7 +689,386 @@ async def _get_dhcp_host_reservations(
...
+            return []
+        if not maas_server_addresses:
+            return []
+        dns_servers = [
+            get_IPvAnyAddress_from_IPAddress(ip)
+            for ip in maas_server_addresses
+            if not ip.is_link_local()
+        ]
+
+        if MAAS_ID.get():
```
AloizioMacedo, line 789: Most of the inside of this `if` seems like it would be better in a separate auxiliary function returning `regions` and `alternate_ips`. The `for address in alternate_ips` loop could stay outside to prevent side effects on the function

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -654,7 +689,386 @@ async def _get_dhcp_host_reservations(
...
+                                    ]
+                                ),
+                            ]
+                        ),
+                        order_by=[
+                            OrderByClause(column=StaticIPAddressTable.c.ip)
+                        ],
+                    )
+                )
+                for region_ip in region_ips:
```
AloizioMacedo: There is a lot of nesting here. Since most of the ifs are just to guard against `None`s and empty things, I think early continues makes it more readable:

```python
                for region_ip in region_ips:
                    if not region_ip.ip:
                        continue

                    interfaces = (
                        await svc.interfaces.get_for_ip(region_ip)
                    ) or []
                    for iface in interfaces:
                        if not iface.node_config_id:
                            continue

                        region_node = await svc.nodes.get_one(
                            query=QuerySpec(
                                where=NodeClauseFactory.with_node_config_id(
                                    iface.node_config_id
                                )
                            )
                        )
                        if not region_node:
                            continue

                        id_plus_family = (
                            region_node.system_id,
                            region_ip.ip.version,
                        )
                        if id_plus_family not in regions:
                            regions.add(id_plus_family)
                            alternate_ips.append(region_ip.ip)
```

I removed the nesting of the `if interfaces` by having the `or []` there as well.

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -654,7 +689,386 @@ async def _get_dhcp_host_reservations(
...
+                    if region_ip.ip:
+                        interfaces = await svc.interfaces.get_for_ip(region_ip)
```
AloizioMacedo: This can be parallelized across region_ips instead of iterating sequentially. For example, you could extract what is inside this if, create tasks for it and then `asyncio.gather` at the end.

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -654,7 +689,386 @@ async def _get_dhcp_host_reservations(
...
+                        if interfaces:
+                            for iface in interfaces:
+                                if iface.node_config_id:
+                                    region_node = await svc.nodes.get_one(
+                                        query=QuerySpec(
+                                            where=NodeClauseFactory.with_node_config_id(
+                                                iface.node_config_id
+                                            )
+                                        )
+                                    )
```
AloizioMacedo: Might be worth it to `get_many` with all `node_config_id`s from the interfaces and filter through it with Python to reduce visits to db.

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -654,7 +689,386 @@ async def _get_dhcp_host_reservations(
...
+                if address not in dns_servers and not address.is_loopback:
+                    dns_servers.append(address)
+        if use_rack_proxy:
+            rack_ips = await self._get_boot_rack_controller_ips(subnet, svc)
+            if dns_servers:
+                dns_servers = rack_ips + [
+                    server
+                    for server in dns_servers
+                    if server not in rack_ips and server != default_region_ip
+                ]
```
AloizioMacedo, line 803: Is there a specific reason to prepend? If not, I would invert the order. (i.e., `filtered_servers+rack_ips`)

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -654,7 +689,386 @@ async def _get_dhcp_host_reservations(
...
+            else:
+                dns_servers = rack_ips
+        if default_region_ip in dns_servers:
+            # Make sure the region DNS server comes last
```
AloizioMacedo, line 807: This comment seems to imply that the order is important. It doesn't seem to be important internally to the function, so it might be externally for some reason.

If so, this should be added to the docstring of the function. If it is not important, the comment should be removed.

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -654,7 +689,386 @@ async def _get_dhcp_host_reservations(
...
+            assert secondary_rack is not None
+            assert secondary_rack.current_config_id is not None
+            node_cfgs.append(secondary_rack.current_config_id)
+        dhcp_vlan_subnet = await svc.subnets.get_one(
+            query=QuerySpec(
+                where=SubnetClauseFactory.with_vlan_id(dhcp_vlan.id)
+            )
+        )
+        assert dhcp_vlan_subnet is not None
+        ips = await svc.staticipaddress.get_for_nodes(
```
AloizioMacedo, line 875: Those asserts should be justified by comment. Without context, it is unclear why they should be true (and they likely depend on how this is being called somewhere else).

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -654,7 +689,386 @@ async def _get_dhcp_host_reservations(
...
+                        StaticIPAddressClauseFactory.with_subnet_id(
+                            dhcp_vlan_subnet.id
+                        ),
+                        NodeClauseFactory.with_node_config_id_in(node_cfgs),
+                    ]
+                )
+            )
+        )
+
+        def rank_ip(ip: IPvAnyAddress):
```
AloizioMacedo, line 892: The logic here of `rank_ip` needs some explanation/references.

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -654,7 +689,386 @@ async def _get_dhcp_host_reservations(
...
+        svc: ServiceCollectionV3,
+        subnet: Subnet,
+        vlan: Vlan,
+        ifaces: list[Interface],
+    ) -> tuple[Interface | None, list[StaticIPAddress] | None]:
+        """Retrieve the best interface IPs that has an IP on the given VLAN.
+        Return it and its IPs.
+
+        Bond interfaces are prefered over physical interfaces, which are prefered over
+        VLAN interfaces.
```
AloizioMacedo: Unclear how ties are solved (or not solved, in which case it would be worth mentioning), if this is a description of the ordering.

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -654,7 +689,386 @@ async def _get_dhcp_host_reservations(
...
+        interfaces = []
+        ifaces_with_static: list[
+            tuple[Interface, list[StaticIPAddress], int]
+        ] = []
+        ifaces_with_discovered: list[
+            tuple[Interface, list[StaticIPAddress], int]
+        ] = []
+        ifaces_on_vlan = [i for i in ifaces if i.vlan_id == subnet.vlan_id]
+        for iface in ifaces_on_vlan:
+            ips = await svc.staticipaddress.get_for_interfaces([iface.id])
```
AloizioMacedo, line 952: I think this could be parallelized similarly to other comments

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -654,7 +689,386 @@ async def _get_dhcp_host_reservations(
...
+                                        ),
+                                    ]
+                                )
+                            )
+                        )
+                        ifaces_with_discovered.append(
+                            (iface, ips, len(dynamic_ranges))
+                        )
+                        break
+        if len(ifaces_with_static) == 1:
```
AloizioMacedo, line 999: This re-assignment of interfaces could be separated into an auxiliary function for clarity, which could itself explain why `lambda t: t[2]` is used as key.

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -654,7 +689,386 @@ async def _get_dhcp_host_reservations(
...
+                    and interface.type == InterfaceType.BOND
+                )
+                or (
+                    best_interface.type == InterfaceType.VLAN
+                    and interface.type == InterfaceType.PHYSICAL
+                )
+            ):
+                best_interface = interface
+                best_interface_ips = interface_ips
+        return best_interface, best_interface_ips
```
AloizioMacedo:
```suggestion
        scores = {
            InterfaceType.VLAN: 0,
            InterfaceType.PHYSICAL: 1,
            InterfaceType.BOND: 2,
        }
        best_interface, best_interface_ips, _ = max(
            interfaces,
            key=lambda i: scores[i[0].type],
            default=(None, None, None),
        )
        return best_interface, best_interface_ips

```

Even better, if this ordering of `InterfaceType` is something that makes sense inherently for it, then implement `__ge__` for it.

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -654,7 +689,386 @@ async def _get_dhcp_host_reservations(
...
+        )
+
+        addr_info = []
+        for ip in rack_addresses:
+            assert ip.subnet_id is not None
+            subnet = await svc.subnets.get_by_id(ip.subnet_id)
+            assert subnet is not None
+            vlan = await svc.vlans.get_by_id(subnet.vlan_id)
+            assert vlan is not None
+            assert ip.ip is not None
```
AloizioMacedo, line 1072: These asserts seem to rely on business logic about the rack addresses. 

We could just `continue` in the loop if these are not true instead (to prevent that one violation stops the whole thing), or at least write out the underlying reasons for the assumptions, like

```python
    assert ip.subnet_id is not None, "rack addresses should always have a subnet_id. This should be enforced by ~this and that~"
```

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -722,22 +1170,96 @@ async def get_dhcp_data_for_agent(
...
+                    else []
+                )
+                next_server = ""
+                for i in ips:
+                    if i.ip and i.ip in subnet.cidr:
+                        next_server = str(i.ip)
+                        break
+                else:
+                    if ips:
+                        next_server = str(ips[0].ip)
```
AloizioMacedo:
```suggestion
                next_server = next(
                    (str(i.ip) for i in ips if i.ip and i.ip in subnet.cidr),
                    str(ips[0].ip) if (ips and ips[0].ip) else "",
                )
```

I think your version also might have a bug, since if `ips` is not empty but `i.ip`  is None for all ips (particularly for the first), for example, then `next_server` will be set to "None" (the string "None", not `None` the value). If this is intended for some reason, it is relying on a very specific quirk. If it is not intended but the scenario I said is impossible, it should be explained.

(These subtle bugs/behaviors are one reason why those imperative approaches can usually be footguns, since it is usual to get lost on how to update things and not on the intent itself.)

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -784,36 +1296,152 @@ async def get_dhcp_data_for_agent(
...
                 host_reservations=hosts,
-                ntp_servers=ntp_servers,
-                default_dns_servers=dns_servers,
+                ntp_servers=global_ntp_servers,
+                default_dns_servers=default_dns_servers,
             )
 
+    async def get_kea_shared_networks_config_ipv4(
+        self, data: DHCPDataForAgent, rack_ip: str
+    ) -> dict[str, Any]:
```
AloizioMacedo, line 1323: Let's add a link to some documentation reference for the kea config in the docstrings of those functions that create them.

AloizioMacedo (changes requested):
Finished with the comments, sorry for not finishing yesterday!

src/maastemporalworker/workflow/dhcp.py
wyattrees: Within this `if` block we are adding elements to `alternate_ips` and `regions` (and subsequent additions to these depend on the current contents of `regions`), so I think parallelizing this actually makes it quite a bit messier for minimal performance gain.

src/maastemporalworker/workflow/dhcp.py
wyattrees: Just following the logic of what's currently in `maasserver/dhcp.py`. I'm assuming that the order chosen there was intentional.

src/maastemporalworker/workflow/dhcp.py
wyattrees: I'm adding a docstring describing the importance of the order.

src/maastemporalworker/workflow/dhcp.py
wyattrees: For the asserts that are working on the assumption that we don't have broken foreign keys in the DB, do you think a comment is really necessary? e.g. if `primary_rack = await svc.nodes.get_by_id(dhcp_vlan.primary_rack_id)` returned `None`, we'd have a VLAN with a `primary_rack_id` for a rack that doesn't actually exist, which should never happen. I agree that comments could be helpful in other cases.

src/maastemporalworker/workflow/dhcp.py
wyattrees: This requires using the `AsyncConnection` object in parallel, which can lead to deadlocks. We could alternatively create a new database transaction for each parallel operation, but I think the overhead of that makes it not worth it.

src/maastemporalworker/workflow/dhcp.py
wyattrees: I don't think this ordering makes sense outside of this context, but I like the `max` implementation you've got.

src/maastemporalworker/workflow/dhcp.py
wyattrees: I added a comment explaining the key.

src/maastemporalworker/workflow/dhcp.py
wyattrees: These asserts are really just to make the linter happy. Some of these asserts are guaranteed by the clauses included in the query above (e.g. `StaticIPAddressClauseFactory.with_subnet_id_not_null()` guarantees that all IPs returned will have `ip.subnet_id != None`), while some are guaranteed by non-nullable foreign keys. I've added comments justifying them.

wyattrees:
@AloizioMacedo all comments have either been resolved or need further discussion

src/maastemporalworker/workflow/dhcp.py
AloizioMacedo: > if primary_rack = await svc.nodes.get_by_id(dhcp_vlan.primary_rack_id) returned None, we'd have a VLAN with a primary_rack_id for a rack that doesn't actually exist, which should never happen.

That is an assumption that is not cohesively contained if it has to be understood here. For instance, another assumption is that for some reason ~the vlan here has to have a primary_rack_id, but it doesn't have to have a secondary_rack_id, even though both are nullable. If I go to the Vlan model (or the Subnet, or Node), there is no explanation why that is the case.~ the primary rack has to have a current_config_id, which isn't clear why and doesn't seem to be a result of a query constraint.

More practically as well, there are cases where we think some validation will hold true but they simply don't. This happened for example in [LP:2147570](https://bugs.launchpad.net/maas/+bug/2147570), where specification of certain things in some ways (with empty strings) were resulting in state that should be invalid.  In this scenario, the asserts would make the code panic with a very opaque error.

We can defer addressing this, but the point is that those assumptions seem to be inherent from the models and from an understanding of MAAS that is not transparent from the code. Ideally, it should reside in an abstraction layer which is reusable and cohesively contained near the models themselves, if they are intrinsic from them. (E.g., maybe a function `get_dhcp_vlan` that takes a subnet and returns either None or the Vlan, and then `get_racks_from_vlan` that returns a list of nodes.)

wyattrees: okay, that makes sense. I've reduced the number of assertions and added context about why the remaining ones should be there.

AloizioMacedo (approved):
Thanks for the quick resolutions, @wyattrees ! :rocket:

PR #358, opened 2026-06-12
feat: register OperationUUID Temporal search attribute

tanzin8r:
Register the OperationUUID keyword search attribute in the Temporal namespace at region worker startup, enabling operation workflows to be linked to operation records.

### DIFF
--- a/src/maastemporalworker/temporal_script.py
+++ b/src/maastemporalworker/temporal_script.py
@@ -7,6 +7,11 @@ import signal
 
 import structlog
 from temporalio.api.enums.v1 import IndexedValueType
+from temporalio.api.operatorservice.v1 import (
+    AddSearchAttributesRequest,
+    ListSearchAttributesRequest,
+)
+from temporalio.client import Client
 
 from maasapiserver.settings import read_config
 from maascommon.worker import set_max_workers_count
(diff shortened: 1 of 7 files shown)


maas-lander:
- [x] Direct message
- [ ] ~maas

src/maastemporalworker/workflow/operation.py
```diff
@@ -0,0 +1,113 @@
...
+    ) -> None:
+        async with self.start_transaction() as services:
+            await services.operations.update_status(
+                operation_uuid=param.operation_uuid,
+                status=param.status,
+                error=param.error,
+            )
+
+
+def track_operation_status(func):
```
r00ta, line 51: why would you blindly track every single activity of the workflow? You would duplicate everything, leading to the double of the db workload. 

Also, you would lose the granularity that only the operation itself should control.

r00ta: never mind, I asked in the other PR

src/maastemporalworker/temporal_script.py
r00ta, line 329: Would it be better to register the search attributes after we setup the namespace here?

r00ta (commented):
inline

src/maastemporalworker/temporal_script.py
tanzin8r: I did think about putting it here. I kept it in Worker because that's already where we create/describe the namespace, and search attributes are tied to that same namespace. Only the shared region worker passes `OperationUUID` anyway. I don't have a strong preference.

r00ta: I think we create the namespace at line where I've put the comment. I'd setup search attributes here as well

r00ta (approved):

PR #360, opened 2026-06-12
fix: fix offline docs deps and packaging

AloizioMacedo:
This PR is intended to solve current issues with the building/bundling of our offline docs.

Those are:
1 - Build issues related to installing the pip dependencies of the offline docs when creating snaps in `s390x` and `ppc64el`. These are primarily issues since some of them (namely, `temporalio`, `pillow`, `bcrypt` and `cryptography`) are mixed with Rust implementations using pyo3, requiring linking with some system packages that may not be there.
2 - The deb version trying to pip install things, even though we do not have internet access during the deb environment.

For 1., the solution here is to patch some dependencies (that we don't really need, and are there just due to the fact that we are extracting the API reference and the CLI reference by walking through the code) or to install the required system dependencies. Of notice is pillow's system dependencies, since pillow is coming from requirements of the documentation itself (not transitive requirements due to the API/CLI reference process).

For 2., we essentially reproduce what we do for the UI and go packages, as they face the same problems.

---

The intent here is to maintain the simplicity, conceptual independence and not rely on external assets, which would not transparently reflect the fact that the documentation is closely tied to the source code and its versioning. For instance, one alternative that was considered was to download the documentation from RTD as offline docs. However:
- The offline documentation is just a sphinx-built html bundle. RTD is a hosting platform, so there is no conceptual reason to couple both. (Only a practical one.)
- There is a chicken-and-egg problem that, for example, in order to build the snap for MAAS 3.X.Y, we would need the documentation of 3.X.Y to be up in RTD, which would mean among other things that 3.X.Y is already tagged. If, for some reason, our build process were to be automated in such a way that pushing a tagged commit would build a snap, this would make building the snap even have potential race conditions.
- It introduces dependency with an external system that we use but do not really control (RTD)

---

Having said the above, it is worth mentioning that the long-term intention here is to make a lot of what exists at the moment unnecessary in the future. From the MAAS side, we should with time be converting to "deprecate" the CLI reference and API reference, having even better alternatives. This solves 80% of the issues, and will happen with a new CLI and the fact that we are using FastAPI for v3 which natively gives a Swagger UI docs page.

The other 20% will be solved by communicating with the Documentation team and making sure that there is a good plan to make the documentation more prone to having LTS, which would likely include things such as not relying on pip-installing the dependencies for it.


maas-lander:
- [ ] Direct message
- [ ] ~maas

alemar99 (approved):
+1

PR #361, opened 2026-06-12
fix: add block-devices and hardware-observe snap plugs required by host-info using LXD 6.8

r00ta:
host-info is using LXD 6.8 which requires access to additional plugs in order to read the machine information properly for the controllers


maas-lander:
- [ ] Direct message
- [ ] ~maas

alexsander-souza (approved):

PR #362, opened 2026-06-15
fix(power): don't run power activities for power drivers of type manual

alemar99:
Resolves: LP:2153152
(cherry picked from commit 30bf78d0c187e4a732d4f8da3f325483eefaeb4f)


maas-lander:
- [ ] Direct message
- [ ] ~maas

PR #364, opened 2026-06-15
fix(dhcp): change decimal to octal format for config file permissions LP#2156073

nehjoshi:
When calling `writeConfigFileDeb` we are passing in an octal value, but parsing it as a decimal value, which was incorrect, and was causing the incorrect permission mode (see bug below).

Resolves [LP:2156073](https://bugs.launchpad.net/maas/+bug/2156073)


maas-lander:
- [x] Direct message
- [ ] ~maas

motrilense (approved):
good finding

PR #365, opened 2026-06-15
fix: change users 'is_active' status when oidc provider is enabled/disabled

alemar99:


maas-lander:
- [ ] Direct message
- [ ] ~maas

src/maasservicelayer/services/external_auth.py
```diff
@@ -451,6 +451,17 @@ async def pre_create_hook(self, builder) -> None:
 
         builder.metadata = await self.get_provider_metadata(builder)
 
+    async def _update_connected_users_active_status(
+        self, provider_id: int, is_active: bool
+    ):
+        """Update all the users connected to `provider_id` to be non active."""
```
nehjoshi: Should be:
```py
 """Update the is_active status of all the users connected to `provider_id`."""
```

src/maasapiserver/v3/middlewares/auth.py
```diff
@@ -478,6 +478,17 @@ async def authenticate(
             )
         )
 
+        if user.is_active is False:
```
nehjoshi:
```py
if not user.is_active:
```
could be better, since `is_active` will always exist in the DB

src/maasapiserver/v3/middlewares/auth.py
```diff
@@ -478,6 +478,17 @@ async def authenticate(
...
+            self._clear_oauth_cookies(request)
+            raise ForbiddenException(
+                details=[
+                    BaseExceptionDetail(
+                        type=INVALID_TOKEN_VIOLATION_TYPE,
```
nehjoshi: Maybe this exception is slightly misleading, since, if the request reaches this part, it proves the tokens are valid, but the user is not active. Something like "NotAuthenticatedViolation" or "ConflictViolation" might be more appropriate

nehjoshi (changes requested):
Looks good overall, with some nitpicks inline

nehjoshi (approved):
LGTM!

PR #367, opened 2026-06-15
feat: add GET handler for operation tasks

MaikRe:


maas-lander:
- [x] Direct message
- [ ] ~maas

src/maasapiserver/v3/api/public/handlers/operations.py
```diff
@@ -0,0 +1,200 @@
...
+                "model": OperationsListResponse,
+            },
+        },
+        response_model_exclude_none=True,
+        status_code=200,
+        dependencies=[Depends(check_authentication())],
+    )
+    async def list_operations(
+        self,
+        filters: OperationFilterParams = Depends(),  # noqa: B008
```
AloizioMacedo: We can avoid having a ton of #noqas in the code by having

`filters: Annotated[OperationFilterParams, Depends()]` 

And similarly in other places. There is explicit guidance in the FastAPI docs to use the annotated version instead of this one: 

Image

---

This didn't exist in previous versions of FastAPI (and Python) since [Annotated was not a thing before 3.9](https://docs.python.org/3/library/typing.html#typing.Annotated), but we should be using this now instead when possible.

src/maasservicelayer/models/operations.py
```diff
@@ -25,3 +33,20 @@ class Operation(MaasTimestampedBaseModel):
...
+    name: str
+    status: OperationTaskStatus
+    result_errors: dict | None = None
+    task_number: int
+    operation_uuid: str
+
+
+class MachineOperationData(BaseModel):
+    op_type: Literal[
+        OperationType.MACHINE_COMMISSION, OperationType.MACHINE_DEPLOY
```
MaikRe: Due to the nature of operation types covering all sorts of operations (not just machine operations) this additional model is required which introduces a little bit of a weird maintainence issue since this list needs to be updated every time `OperationType` receives a new machine related type. Happy for any comments/recommendations on how to improve this.

src/maasapiserver/v3/api/public/handlers/operations.py
```diff
@@ -128,7 +132,71 @@ async def get_operation(
...
+    ) -> OperationTasksListResponse:
+        """Get a paginated list of tasks for a specific operation."""
+
+        assert authenticated_user is not None
+        can_view_all = (
+            await services.openfga_tuples.get_client().can_view_operations(
+                authenticated_user.id
+            )
+        )
+        await services.operations.get_by_uuid_for_user(
```
alemar99, line 172: This is possibly None if the user doesn't have access to the operation or the operation doesn't exist, and so you should raise a NotFoundException

src/maasapiserver/v3/api/public/models/responses/operations.py
alemar99: This should be just `result`. In the spec it was written as `result/errors` to make it clear it contained both.  I didn't notice in the previous PRs. You don't have to correct it in this PR, you can do it later.

src/maasapiserver/v3/api/public/models/responses/operations.py
```diff
@@ -64,5 +81,45 @@ def from_model(
...
 
 
+class OperationTaskResponse(HalResponse[BaseHal]):
+    kind: str = Field(default="OperationTask")
+    id: int
+    started_at: datetime | None = None
+    finished_at: datetime | None = None
+    name: str
+    status: OperationTaskStatus
+    result_errors: dict | None = None
```
alemar99: result

src/maasapiserver/v3/api/public/models/responses/operations.py
```diff
@@ -12,8 +12,22 @@
...
+)
+from maasservicelayer.models.operations import (
+    MachineOperationData,
+    Operation,
+    OperationTask,
+)
+
+TypeSpecificData = Annotated[
+    MachineOperationData,
+    Field(discriminator="op_type"),
```
alemar99: I remember it was not ideal to have the discriminator due to some issue with the generated openapi spec. Can't remember the details now

src/maasapiserver/v3/api/public/models/responses/operations.py
```diff
@@ -33,12 +47,14 @@ class OperationResponse(HalResponse[BaseHal]):
     is_bulk: bool
     parent_id: str | None = None
     user_id: int | None = None
+    type_specific_data: TypeSpecificData | None = None
```
alemar99: Maybe have a `MachineOperationResponse` which inherits from `OperationResponse` and adds the missing fields?

alemar99 (changes requested):
First bunch of comments, I have to come back at it tomorrow

src/maasservicelayer/models/operations.py
AloizioMacedo: Perhaps something like

```python
class MachineOperation(StrEnum):
    MACHINE_COMMISSION = "machine.commission"
    MACHINE_DEPLOY = "machine.deploy"
    MACHINE_BULKDEPLOY = "machine.bulkdeploy"


class OperationType(StrEnum):
    MACHINE_COMMISSION = MachineOperation.MACHINE_COMMISSION
    MACHINE_DEPLOY = MachineOperation.MACHINE_DEPLOY
    MACHINE_BULKDEPLOY = MachineOperation.MACHINE_BULKDEPLOY
    SELECTION_SYNC = "selection.sync"
```

in `enums/operations.py`? A little bit of duplication, but more cohesive so easier to maintain. I don't think there is anything much cleaner, as Python is quite limited in terms of enums.

Then `MachineOperationData` would be 

```python
class MachineOperationData(BaseModel):
    op_type: MachineOperation
```

That said, isn't `MachineOperationData` missing the bulkdeploy? Or is it not supposed to be there?

src/maasservicelayer/services/operations.py
```diff
@@ -111,3 +125,34 @@ async def get_by_uuid_for_user(
...
+            size=size,
+        )
+
+    async def get_type_specific_data(
+        self, operation: Operation
+    ) -> MachineOperationData | None:
+        match operation.op_type:
+            case (
+                OperationType.MACHINE_COMMISSION | OperationType.MACHINE_DEPLOY
+            ):
```
AloizioMacedo: If going with the suggestion from the comment about the enums, this could likely change to 

```python
if operation in MachineOperation: 
    ...
```

AloizioMacedo: Again depending on whether BULK_DEPLOY should be involved or not.

src/maasservicelayer/models/operations.py
MaikRe: Thank you for the suggestion, I will have a look at how that would refactor the code.


> That said, isn't `MachineOperationData` missing the bulkdeploy? Or is it not supposed to be there?

I am not sure about this yet because each Machine Operation is tied to a specific node so really it could only be a parent operation that contains this type and not a machine operation.
Maybe @alemar99 can comment on this.

src/maasapiserver/v3/api/public/models/responses/operations.py
MaikRe: Handling this here which should optimally be merged before this PR:
https://github.com/canonical/maas/pull/378

src/maasapiserver/v3/api/public/models/responses/operations.py
MaikRe: The problem I see with that is that the GET /operations endpoint which lists all operations would either need to return a list which potentially mixes responses as part of the response list which I am not sure is reasonable, or it must only return a list of one time in order to have something like MachineOperationsListResponse. Nesting the type specific data within the OperationResponse ensures that the API level types are consistent

src/maasservicelayer/models/operations.py
alemar99: Yes, I agree with Maik. The operation is not tied to a specific node but spawns children operations which are related to a single node

alemar99:
I thought on this, and maybe it doesn't make sense for the time being to add the complexity of machine-specific fields for operations (since now the only field being added is the `node_id`, which is also redundant with the `resource_id` already in the operation). I'd say that we should remove that part and focus only on the operations and operation tasks. Once we have things clear on what data we want to include for this kind of operation, then we'll think on how to implement this.
I would not like to jump into an implementation now for some use cases which we don't have clear in our mind, possibly having to revert this in the future.

MaikRe:
> I thought on this, and maybe it doesn't make sense for the time being to add the complexity of machine-specific fields for operations (since now the only field being added is the `node_id`, which is also redundant with the `resource_id` already in the operation). I'd say that we should remove that part and focus only on the operations and operation tasks. Once we have things clear on what data we want to include for this kind of operation, then we'll think on how to implement this. I would not like to jump into an implementation now for some use cases which we don't have clear in our mind, possibly having to revert this in the future.

Do you think its worth cleaning up the migration files as database tables for this as well then? @alemar99

alemar99:
@MaikRe Yes, I think it makes sense to revert any reference to them.
To keep things even clearer, we could have done the work in a feature branch, but since we already started on master let's continue here.

src/maasapiserver/v3/api/public/handlers/operations.py
```diff
@@ -130,12 +133,75 @@ async def get_operation(
             user_id=authenticated_user.id,
             can_view_all=can_view_all,
         )
+        if operation is None:
+            raise NotFoundException()
```
alemar99: This is dead code as the `get_by_uuid_for_user` either returns an Operation or raises an exception.

src/tests/maasservicelayer/db/repositories/test_operations.py
```diff
@@ -298,31 +299,6 @@ async def test_update_many(
...
-                "is_bulk": is_bulk,
-                "created": now,
-                "updated": now,
-            }
-        ],
-    )
-    return Operation(**row)
-
-
 async def create_test_operation_task_entry(
```
alemar99: Did you forgot to delete this as well?

src/maasapiserver/v3/api/public/handlers/operations.py
```diff
@@ -130,12 +133,75 @@ async def get_operation(
...
+                authenticated_user.id
+            )
+        )
+        operation = await services.operations.get_by_uuid_for_user(
+            operation_uuid,
+            user_id=authenticated_user.id,
+            can_view_all=can_view_all,
+        )
+        if operation is None:
+            raise NotFoundException()
```
alemar99: Dead code

alemar99 (dismissed):
Approving with a comment: having the `get_by_uuid_for_user` automatically raising the `NotFoundException` makes things a little bit misleading in the handler code (i.e. why we get an operation if we not use that later?).
IMHO we should return None if the operation doesn't exists. We can address it later.

alemar99 (approved):

PR #368, opened 2026-06-15
feat: add operation task tracking via local activity

tanzin8r:


maas-lander:
- [x] Direct message
- [ ] ~maas

PR #369, opened 2026-06-15
feat: add operation task tracking via local activity

tanzin8r:
Workflows can call `update_current_task` before each step to persist the current task on the operation and create a matching task row. Also, clear current_task when the operation completes successfully.


maas-lander:
- [x] Direct message
- [ ] ~maas

src/maastemporalworker/workflow/operation.py
```diff
@@ -47,24 +56,60 @@ async def update_operation_status(
...
+def _get_operation_uuid() -> str:
+    """Return the operation UUID tracked by the running workflow.
+
+    The UUID is read from the ``OperationUUID`` search attribute set when the
+    workflow is started. Raises if the attribute is missing.
+    """
+    info = workflow.info()
+    operation_uuid = info.search_attributes.get(
+        OPERATION_UUID_SEARCH_ATTRIBUTE
+    )
```
AloizioMacedo: `info.search_attributes` seems to be deprecated::

Image

And changing it seems to kill two birds with one stone, since you get rid of the awkward operation_uuid[0].

You can likely declare OPERATION_UUID_SEARCH_ATTRIBUTE as follows:

```python
from temporalio.common import SearchAttributeKey

OPERATION_UUID_SEARCH_ATTRIBUTE = SearchAttributeKey.for_keyword(
    "OperationUUID"
)
``` 

And then

```python
def _get_operation_uuid() -> str:
    """Return the operation UUID tracked by the running workflow.

    The UUID is read from the ``OperationUUID`` search attribute set when the
    workflow is started. Raises if the attribute is missing.
    """
    info = workflow.info()
    operation_uuid = info.typed_search_attributes.get(
        OPERATION_UUID_SEARCH_ATTRIBUTE
    )
    if not operation_uuid:
        raise ApplicationError(
            f"Operation tracking is enabled for workflow {info.workflow_type}"
            f" but the search attribute {OPERATION_UUID_SEARCH_ATTRIBUTE}"
            " has not been set."
        )

    return operation_uuid
```

src/maasservicelayer/services/operations.py
```diff
@@ -47,6 +51,10 @@ async def update_status(
             OperationStatus.FAILED,
         ):
             builder.finished = utcnow()
+        # On success the operation is no longer running any task; on failure we
+        # leave current_task untouched so the user can see where it stopped.
+        if status == OperationStatus.COMPLETED:
+            builder.current_task = ""
```
AloizioMacedo: Why not 

```suggestion
            builder.current_task = None
```
?

tanzin8r: I used an empty string intentionally here, as the [spec](https://docs.google.com/document/d/1vqGR7Z18acP7SpwaiVcXPBXUFr06HNP57n0VAWHSKzY/edit?tab=t.0#heading=h.frfjj7jg2okz)  specifies clearing `current_task` with an empty string on successful run. I believe `NULL` is simply the unset state before any task name is recorded. tagging @alemar99 so we can get more insights into this.

alemar99: Since `current_task` is nullable it makes sense to have it set to `None`

src/maastemporalworker/workflow/operation.py
```diff
@@ -47,24 +60,59 @@ async def update_operation_status(
...
+    workflow is started. Raises if the attribute is missing.
+    """
+    info = workflow.info()
+    operation_uuid = info.typed_search_attributes.get(
+        OPERATION_UUID_SEARCH_ATTRIBUTE
+    )
+    if not operation_uuid:
+        raise ApplicationError(
+            f"Operation tracking is enabled for workflow {info.workflow_type}"
+            f" but the search attribute {OPERATION_UUID_SEARCH_ATTRIBUTE}"
```
AloizioMacedo:
```suggestion
            f" but the search attribute {OPERATION_UUID_SEARCH_ATTRIBUTE.name}"
```

Sorry, @tanzin8r ! Forgot to change this part in my suggestion. Since it is not a string anymore, it is better to use the name in the error message (see below)

--- 

This:

```python
from temporalio.common import SearchAttributeKey

OPERATION_UUID_SEARCH_ATTRIBUTE = SearchAttributeKey.for_keyword("OperationUUID")

print(OPERATION_UUID_SEARCH_ATTRIBUTE)
print("---")
print(OPERATION_UUID_SEARCH_ATTRIBUTE.name)
```

prints this:

```bash
_SearchAttributeKey(_name='OperationUUID', _indexed_value_type=<SearchAttributeIndexedValueType.KEYWORD: 2>, _value_type=<class 'str'>)
---
OperationUUID
```

tanzin8r: My bad. I overlooked this as well. Since it's a key object now.

src/maasservicelayer/services/operations.py
```diff
@@ -111,3 +129,31 @@ async def get_by_uuid_for_user(
...
+            builder=OperationTaskBuilder(
+                operation_uuid=operation_uuid,
+                name=name,
+                task_number=task_number,
+                status=OperationTaskStatus.RUNNING,
+                started_at=utcnow(),
+            )
+        )
+        await self.operations_service.set_current_task(operation_uuid, name)
+        return task
```
AloizioMacedo: One last question, perhaps to @alemar99 as well.

Is there some design decision that imposes the creation of this separate service? E.g., does it **have** to exist because there is a table associated to the tasks?

I ask because it looks a lot like an unnecessary [middle man](https://refactoring.guru/smells/middle-man), at least at this point. It just calls the operations service to do what it needs.

tanzin8r: I don't think it have to exist by design. I can move `start_task` to `OperationsService`.

AloizioMacedo: OK! @tanzin8r Let's do that and then we can move forward from my POV : )

tanzin8r: Just changed it. you can take a look again

AloizioMacedo (approved):

PR #370, opened 2026-06-16
fix: move the block-devices snap plug under pebble

alemar99:
`/usr/sbin/blkid` access was denied because the correct profile to apply the plug is pebble and not maas.

Dmesg apparmor message:
``` 
audit: type=1400 audit(1781602195.196:6720172): apparmor="DENIED" operation="exec" class="file" namespace="root//lxd-foo_<var-snap-lxd-common-lxd>" profile="snap.maas.pebble" name="/usr/sbin/blkid" pid=2093677 comm="amd64" requested_mask="x" denied_mask="x" fsuid=1000000 ouid=1000000
```

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -50,6 +50,7 @@ apps:
     command: usr/bin/run-pebble-daemon
     plugs:
       - avahi-observe
+      - block-devices  # allow blkid (machine-resources) to read disk devices
       - hardware-observe
       - kernel-module-observe
       - mount-observe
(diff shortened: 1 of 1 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

snap/snapcraft.yaml
```diff
@@ -65,8 +66,6 @@ apps:
   maas:
     command: usr/bin/maas
     plugs:
-      - block-devices  # allow blkid (machine-resources) to read disk devices
-      - hardware-observe  # allow machine-resources to read hardware info from /sys
```
r00ta, line 69: you need hardware-observe otherwise the host-info can't read information of gpu, at least on my system

alemar99: hardware-observe plug is already there for pebble

r00ta (approved):

PR #371, opened 2026-06-16
fix(deps): include missing python3 dependencies for debs

alemar99:
The python3-ulid and python3-tz were missing in the debian control file, causing maas not being able to be installed/executed.


maas-lander:
- [ ] Direct message
- [ ] ~maas

alexsander-souza (approved):

PR #373, opened 2026-06-16
fix(dhcp): change decimal to octal format for config file permissions LP#2156073 (#364)

nehjoshi:
When calling `writeConfigFileDeb` we are passing in an octal value, but parsing it as a decimal value, which was incorrect, and was causing the incorrect permission mode (see bug link).

Resolves: [LP:2156073](https://bugs.launchpad.net/maas/+bug/2156073)
(cherry picked from commit [cd37a9d](https://github.com/canonical/maas/commit/cd37a9d65d708103101057d107859124bca3f2aa))


maas-lander:
- [ ] Direct message
- [ ] ~maas

PR #377, opened 2026-06-16
fix: sudo-rs doesn't support "-E" flag

alemar99:
Replace the occurrences with explicit `--preserve-env=<var>` where those vars were known.


maas-lander:
- [ ] Direct message
- [ ] ~maas

alemar99:
This resulted in scripts result to be like this:
```bash
ubuntu@debmaas:~$ cat maas-commission-9jfr_9t1/out/50-maas-01-commissioning
sudo: preserving the entire environment is not supported, '-E' is ignored
ubuntu@debmaas:~$ cat maas-commission-9jfr_9t1/out/20-maas-03-machine-resources
sudo: preserving the entire environment is not supported, '-E' is ignored
Prepended http:// to '/MAAS/machine-resources/amd64'
http:///MAAS/machine-resources/amd64: Invalid host name.
```

r00ta (commented):
is this a problem also when resolute is set used as commissioning image? We might have to backport to 3.7, 3.6 and 3.5

alemar99:
Good point, I didn't tested but I suppose that's the case

src/metadataserver/user_data/templates/snippets/maas_run_remote_scripts.py
```diff
@@ -167,11 +167,33 @@ def download_and_extract_tar(url, creds, scripts_dir):
         return True
 
 
+# sudo-rs doesn't implement the -E flag, so we pass --preserve-env for each
+# environment variable we need to keep. These cover the proxy configuration
+# and the apt non-interactive frontend.
+# XXX: Some env vars previously passed might not be included in this list.
+SUDO_PRESERVE_ENV_VARS = [
```
tmerten, line 174: Not sure if we usually do this as a practice but this could be reused in the test.

src/provisioningserver/refresh/__init__.py
```diff
@@ -121,8 +121,21 @@ def runscripts(
             "RESULT_PATH": result_path,
             "TMPDIR": tmpdir,
         }
+
+        env = os.environ | env_vars
```
tmerten: nitpick, this is not needed here, can be moved closer to or possibly in try block below

tmerten (changes requested):
a question and a nitpick inline

tmerten (approved):
lovely, ty

PR #379, opened 2026-06-16
feat: add delete operations handler for gracefully handling operation cancellation

MaikRe:


maas-lander:
- [x] Direct message
- [ ] ~maas

src/maasapiserver/v3/api/public/handlers/operations.py
```diff
@@ -132,3 +135,49 @@ async def get_operation(
...
+        authenticated_user: Annotated[
+            AuthenticatedUser | None, Depends(get_authenticated_user)
+        ],
+        services: Annotated[ServiceCollectionV3, Depends(services)],
+    ) -> OperationResponse:
+        """Cancel a specific operation by UUID."""
+
+        assert authenticated_user is not None
+        can_view_all = (
+            await services.openfga_tuples.get_client().can_view_operations(
```
alemar99, line 171: It should be `can_edit_operations`

src/maastemporalworker/workflow/operation.py
```diff
@@ -94,11 +94,15 @@ async def wrapper(self, param):
                 operation_uuid=operation_uuid,
                 error=str(e),
             )
+            if isinstance(e, CancelledError):
```
alemar99: There is `is_cancelled_exception` from `temporalio.exceptions`

src/maastemporalworker/workflow/operation.py
```diff
@@ -94,11 +94,15 @@ async def wrapper(self, param):
...
+                status = OperationStatus.CANCELLED
+            else:
+                status = OperationStatus.FAILED
             await workflow.execute_local_activity(
                 UPDATE_OPERATION_STATUS_ACTIVITY_NAME,
                 UpdateOperationStatusParam(
                     operation_uuid=operation_uuid,
-                    status=OperationStatus.FAILED,
+                    status=status,
                     error=str(e),
```
alemar99: I'd say to write an error message when the operation is cancelled, rather than returning the str representation of the exception

src/tests/maasservicelayer/services/test_temporal.py
```diff
@@ -206,3 +209,74 @@ async def test_workflow_status(
...
+        handle.cancel.assert_awaited_once()
+
+    async def test_cancel_workflow_by_operation_uuid_no_running_workflows(
+        self,
+        service: TemporalService,
+        temporal_client_mock,
+    ) -> None:
+        async def _list_workflows(query):
+            return
+            yield  # noqa: unreachable - makes this an async generator
```
alemar99: `yield from ()` could be an option?

alemar99 (changes requested):
Some nits inline

src/tests/maasservicelayer/services/test_temporal.py
MaikRe: @alemar99 yield from is not allowed in the async function and therefor isn't an option.
Alternative could be something like
```
async def _list_workflows(query):
    for _ in ():
        yield
```

src/maasservicelayer/services/operations.py
```diff
@@ -111,3 +113,40 @@ async def get_by_uuid_for_user(
...
+        self,
+        uuid: str,
+        user_id: int,
+        can_edit_all: bool,
+    ) -> Operation:
+        operation = await self.get_by_uuid_for_user(
+            uuid, user_id=user_id, can_view_all=can_edit_all
+        )
+        if operation is None:
+            raise NotFoundException(
```
alemar99: I believe this is not 100% correct. We should raise a NotFoundException if the user can't see the operation or the operation doesn't exist.

Otherwise, we should raise a Forbidden exception if the user can't edit all and the operation.user_id != user.id

src/tests/maasservicelayer/services/test_temporal.py
alemar99: looks good

src/maasservicelayer/services/operations.py
MaikRe: I forgot this is already handled in the get_by_uuid_for_user function, removed this block.

src/maasservicelayer/services/operations.py
```diff
@@ -111,3 +113,31 @@ async def get_by_uuid_for_user(
...
         return operation
+
+    async def cancel_for_user(
+        self,
+        uuid: str,
+        user_id: int,
+        can_edit_all: bool,
+    ) -> Operation:
+        operation = await self.get_by_uuid_for_user(
+            uuid, user_id=user_id, can_view_all=can_edit_all
```
alemar99: Again you should not mix view and edit permission.
If a user can view but can't edit all the operations, the right response to be returned is a Forbidden (403). Here instead you will return a Not Found (404) which is misleading because the user can actually see the operation.

I think the solution would be to pass both `can_view_all` and `can_edit_all` to the `cancel_for_user`.

Then:
1. call `get_by_uuid_for_user` with `can_view_all`
2. `if not can_edit_all and operation.id != user.id:` -> raise ForbiddenException
3. Your logic

src/tests/maasapiserver/v3/api/public/handlers/test_operations.py
```diff
@@ -193,6 +202,68 @@ async def test_get_operation_not_found(
...
+        services_mock.operations = Mock(OperationsService)
+        services_mock.operations.cancel_for_user.side_effect = (
+            ConflictException()
+        )
+
+        response = await client.delete(
+            f"{self.BASE_PATH}/{TEST_OPERATION.uuid}"
+        )
+        assert response.status_code == 409
+
```
alemar99, line 269: add a test for the use case in my other comment

src/tests/maasservicelayer/services/test_operations.py
```diff
@@ -391,3 +394,78 @@ async def test_get_by_uuid_for_user_not_found(
...
+        operations_repo_mock.get_one.return_value = operation
+        operations_repo_mock.update_by_id.return_value = operation.model_copy(
+            update={"status": OperationStatus.CANCELLING}
+        )
+
+        result = await operations_service.cancel_for_user(
+            uuid="op-uuid", user_id=1, can_edit_all=True
+        )
+
+        assert result.status == OperationStatus.CANCELLING
```
alemar99: These can be merged into one with `pytest.mark.parametrize`

src/tests/maasservicelayer/services/test_operations.py
```diff
@@ -391,3 +394,78 @@ async def test_get_by_uuid_for_user_not_found(
...
+        self,
+        operations_service: OperationsService,
+        operations_repo_mock: Mock,
+    ) -> None:
+        operations_repo_mock.get_by_uuid.return_value = None
+
+        with pytest.raises(NotFoundException):
+            await operations_service.cancel_for_user(
+                uuid="nonexistent-uuid", user_id=1, can_edit_all=True
+            )
```
alemar99, line 478: same, add a test

alemar99 (approved):
LGTM

PR #383, opened 2026-06-17
feat: add adapters for EntraID, Auth0 & Keycloak

nehjoshi:
- Added adapters for MS Entra ID, Auth0 and Keycloak
- Using the client credentials, we first get an access token from the provider. Then, we use the access token to get the status of the user
- Each provider has their own mechanism. See the reference links below
- For now, I'm keeping this in a feature branch in case we decide to deviate significantly. Also, at some point, we will need a migration for adding a "vendor" property to the `OIDCProviderTable`, which will be an IntEnum for each provider we support. 

[Auth0 Client Credentials Flow](https://auth0.com/docs/get-started/authentication-and-authorization-flow/client-credentials-flow)
[Auth0 Management API Reference](https://auth0.com/docs/api/management/v2)
[Entra ID Client Credentials Flow](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-client-creds-grant-flow)
[MS Graph API](https://learn.microsoft.com/en-us/graph/identity-network-access-overview)
[Keycloak Docs](https://www.keycloak.org/docs/latest/)


maas-lander:
- [ ] Direct message
- [ ] ~maas

src/maasservicelayer/auth/oidc_adapters.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class BaseProviderAdapter(ABC):
+    """Queries a provider's user-management API using the client credentials
+    (machine-to-machine) flow, to verify whether a user is still valid.
+    """
+
+    def __init__(self, provider: OAuthProvider, http_client: AsyncClient):
+        self.provider = provider
+        self._http = http_client
```
nehjoshi, line 35: We will reuse the `httpx_client` created in the service layer

alemar99:
@nehjoshi I think we can close this since we have https://github.com/canonical/maas/pull/425 ?

nehjoshi:
Yep, this is an old one

PR #384, opened 2026-06-17
fix: sudo-rs doesn't support "-E" flag

alemar99:
Replace the occurrences with explicit `--preserve-env=<var>` where those vars were known.

(cherry picked from commit 203f8c724db355639f2cc47141b2623453b14a35)


maas-lander:
- [x] Direct message
- [ ] ~maas

AloizioMacedo:
sudo-rs was not being used in noble and before, right? So why is the backport necessary?

alemar99:
> sudo-rs was not being used in noble and before, right? So why is the backport necessary?

It is also used in commissioning scripts which might be executed by machines running resolute

PR #387, opened 2026-06-17
feat(dhcp): Create new maas-kea-dhcp-helper to notify MAAS of lease comits, expirations, ane releases. Add script to Kea configuration.

wyattrees:


maas-lander:
- [ ] Direct message
- [ ] ~maas

package-files/usr/sbin/maas-kea-dhcp-helper
```diff
@@ -0,0 +1,89 @@
...
+        ip = os.environ[f"LEASES4_AT{ln}_ADDRESS"]
+        notify("commit", mac, "ipv4", ip, 
+               hostname=os.getenv(f"LEASES4_AT{ln}_HOSTNAME"),
+               lease_time=os.getenv(f"LEASES4_AT{ln}_VALID_LIFETIME")
+        )
+    return
+
+def lease4_expire():
+    notify(
+        "expire",
```
tanzin8r: I think the action string should be `expiry`.  check LeaseAction `src/maascommon/enums/ipaddress.py`

package-files/usr/sbin/maas-kea-dhcp-helper
```diff
@@ -0,0 +1,89 @@
...
+def leases4_committed():
+    num_leases = os.environ["LEASES4_SIZE"]
+    for ln in range(int(num_leases)):
+        mac = os.environ[f"LEASES4_AT{ln}_HWADDR"]
+        ip = os.environ[f"LEASES4_AT{ln}_ADDRESS"]
+        notify("commit", mac, "ipv4", ip, 
+               hostname=os.getenv(f"LEASES4_AT{ln}_HOSTNAME"),
+               lease_time=os.getenv(f"LEASES4_AT{ln}_VALID_LIFETIME")
+        )
+    return
```
MaikRe: redundant return

package-files/usr/sbin/maas-kea-dhcp-helper
```diff
@@ -0,0 +1,89 @@
...
+"""Utility script used by MAAS's Kea DHCP servers."""
+
+import sys
+import os
+import time
+import json
+import socket
+from contextlib import closing
+
+DHCP_SOCK = os.getenv("MAAS_DHCP_SOCK", "/var/snap/maas/common/maas/dhcpd.sock")
```
MaikRe, line 15: Just flagging that if the env variable does not get properly set/passed for the deb it will fall back to the snap socket path. Unlikely to happen but a potential silent bug

package-files/usr/sbin/maas-kea-dhcp-helper
```diff
@@ -0,0 +1,89 @@
...
+        "lease4_release": lease4_release,
+    }
+    try:
+        hooks[hook]()
+    except KeyError:
+        # several other hooks may be called, but we don't care about them.
+        return
+
+if __name__ == "__main__":
+    main(sys.argv[1])
```
MaikRe: Nit: missing new line at the end of file

MaikRe: If sys.argv[1] is empty this will raise an index error. Primarily a nit since this is just a POC

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -1318,6 +1319,18 @@ async def get_dhcp_data_for_agent(
                 default_dns_servers=default_dns_servers,
             )
 
+    def get_hooks_libraries(self) -> dict[str, Any]:
```
MaikRe:
```suggestion
    def get_kea_hooks_libraries(self) -> dict[str, Any]:
```
For naming consistency?

MaikRe (commented):
Some nits

package-files/usr/sbin/maas-kea-dhcp-helper
wyattrees: This script will inherit the environment of the Kea process, for which we are setting this variable in this PR (for both the deb and snap). So this shouldn't ever happen, and there's not really anything better that we can do as far as I know since we can't detect which type of install we have within this script.

package-files/usr/sbin/maas-kea-dhcp-helper
wyattrees: The script is called by Kea and always includes the name of the hook as the only argument. Should that ever change in Kea we can handle this differently

tanzin8r:
should we also add to `debian/maas-dhcp.apparmor` at this point? will `libdhcp_run_script.so` need an AppArmor rule to exec `maas-kea-dhcp-helper`?

wyattrees:
> should we also add to `debian/maas-dhcp.apparmor` at this point? will `libdhcp_run_script.so` need an AppArmor rule to exec `maas-kea-dhcp-helper`?

This will happen with [MAASENG-6543](https://warthogs.atlassian.net/browse/MAASENG-6543)

[MAASENG-6543]: https://warthogs.atlassian.net/browse/MAASENG-6543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #389, opened 2026-06-18
refactor: filter httpx logging in openfga clients

alemar99:
After introducing openfga, we had many logs related to requests to the openfga store, like:

```
Jun 18 12:35:48 maas-resolute maas-regiond[3398]: httpx: [info] HTTP Request: POST http://unix/stores/00000000000000000000000000/list-objects "HTTP/1.1 200 OK"
Jun 18 12:35:48 maas-resolute maas-regiond[3398]: httpx: [info] HTTP Request: POST http://unix/stores/00000000000000000000000000/check "HTTP/1.1 200 OK"
```

Since they don't bring any additional value, and might actually confuse users, we change the logging level for those `httpx` calls to warning.

### DIFF
--- a/src/maascommon/openfga/async_client.py
+++ b/src/maascommon/openfga/async_client.py
@@ -34,32 +34,34 @@ class OpenFGAClient(BaseOpenFGAClient):
         await self.client.aclose()
 
     async def _check(self, user_id: int, relation: str, obj: str) -> bool:
-        response = await self.client.post(
-            f"/stores/{OPENFGA_STORE_ID}/check",
-            json={
-                "tuple_key": {
-                    "user": f"user:{user_id}",
-                    "relation": relation,
-                    "object": obj,
+        with self._suppress_httpx_logging():
+            response = await self.client.post(
+                f"/stores/{OPENFGA_STORE_ID}/check",
+                json={
+                    "tuple_key": {
+                        "user": f"user:{user_id}",
+                        "relation": relation,
+                        "object": obj,
+                    },
+                    "authorization_model_id": OPENFGA_AUTHORIZATION_MODEL_ID,
                 },
-                "authorization_model_id": OPENFGA_AUTHORIZATION_MODEL_ID,
-            },
-        )
+            )
         response.raise_for_status()
         return response.json().get("allowed", False)
 
     async def _list_objects(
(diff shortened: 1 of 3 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

alemar99:
httpx doesn't support client-level logging, but only module-level, see [their docs](https://www.python-httpx.org/logging/).

The context manager changes the logging only for the http requests to the openfga url and then restores it to what was previously set.

You can test this behavior with the following script:
```python
import asyncio
import logging
from contextlib import contextmanager
import contextvars

import httpx

logging.basicConfig(
    level=logging.INFO,
    format="%(name)-12s %(levelname)-8s %(message)s",
)


_suppress_httpx = contextvars.ContextVar("suppress_httpx", default=False)


class _HTTPXLogFilter(logging.Filter):

    def filter(self, record: logging.LogRecord) -> bool:
        if record.levelno >= logging.WARNING:
            return True
        return not _suppress_httpx.get()


def _install_httpx_log_filter() -> None:
    for name in ("httpx", "httpcore"):
        logging.getLogger(name).addFilter(_HTTPX_LOG_FILTER)

_HTTPX_LOG_FILTER = _HTTPXLogFilter()
_install_httpx_log_filter()


@contextmanager
def suppress_httpx_logging():
    with _suppress_httpx.set(True):
        yield

async def get_req(client, path=""):
    url = "https://example.com" + path
    await client.get(url)


async def main():

    async def no_log_http_call(client):
        with suppress_httpx_logging():
            # add a query param to differentiate from others
            await get_req(client, path="?foo=bar")

    client = httpx.AsyncClient()

    async with asyncio.TaskGroup() as tg:
        tg.create_task(get_req(client))
        tg.create_task(no_log_http_call(client))
        tg.create_task(get_req(client))

    await client.aclose()

if __name__ == "__main__":
    asyncio.run(main())
```

Here I create 3 async tasks that run concurrently, to make sure that the suppression only applies to one. You can change the logging level to `logging.DEBUG` to see also httpcore logs

r00ta (changes requested):
what about this https://github.com/canonical/maas/pull/480?

alemar99:
Closed because of #480

PR #390, opened 2026-06-18
fix: re-arm status_expires when entering disk erasing LP#2155614

skyecas:
nodes in disk_erasing were silently transitioning to failed because the prior lifecycle timeout had expired. the `_start_disk_erasing()` function was not arming a fresh status expires from, the node timeout.

Resolves LP#2155614


maas-lander:
- [ ] Direct message
- [ ] ~maas

r00ta:
The bug references 3.4 and the logic around the release operation has changed in 3.5. Did you manage to reproduce it in master?

PR #395, opened 2026-06-18
feat(dhcp): add host reservations to Kea config

wyattrees:
This also introduces the InterfaceRelationshipTable object in maasservicelayer as a necessity.

### DIFF
--- a/src/maastemporalworker/workflow/dhcp.py
+++ b/src/maastemporalworker/workflow/dhcp.py
@@ -67,7 +67,6 @@ from maastemporalworker.workflow.utils import (
     activity_defn_with_context,
     workflow_run_with_context,
 )
-from provisioningserver.path import get_path
 
 from provisioningserver.boot.pxe import PXEBootMethod  # noqa:E402 isort:skip
 from provisioningserver.boot.grub import UEFIAMD64BootMethod
--- a/src/maastemporalworker/workflow/dhcp.py
+++ b/src/maastemporalworker/workflow/dhcp.py
@@ -67,6 +67,7 @@ from maastemporalworker.workflow.utils import (
     activity_defn_with_context,
     workflow_run_with_context,
 )
+from provisioningserver.path import get_path
 
 from provisioningserver.boot.pxe import PXEBootMethod  # noqa:E402 isort:skip
 from provisioningserver.boot.grub import UEFIAMD64BootMethod
@@ -176,6 +177,13 @@ class IPRangeData:
 
 
 @dataclass
+class HostReservationData:
+    ip: str
+    mac_address: str
+    hostname: str
+
+
+@dataclass
 class SubnetData:
     id: int
     ip_version: int
@@ -192,6 +200,7 @@ class SubnetData:
     ntp_servers: list[str] | None
     next_server: str
     pools: list[IPRangeData]
+    host_reservations: list[HostReservationData]
 
 
 @dataclass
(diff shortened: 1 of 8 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -579,116 +575,85 @@ async def get_active_interfaces_for_agent(
...
+                                mac_address=interface.mac_address,
+                                hostname=hostname,
                             )
                         )
-                    )
-                    assert node is not None
-                    if node.node_type in (
-                        NodeTypeEnum.RACK_CONTROLLER,
-                        NodeTypeEnum.REGION_CONTROLLER,
-                        NodeTypeEnum.REGION_AND_RACK_CONTROLLER,
```
MaikRe, line 655: Removing this means KEA is also responsible for reservations of the racks own interfaces, is that the intended behavior?

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -1094,6 +1059,26 @@ def sort_key__dhcp_on__ip(
         }
         return {key: str(value[3]) for key, value in best_ntp_servers.items()}
 
+    async def _make_interface_hostname(
+        self, interface: Interface, svc: ServiceCollectionV3
+    ) -> str:
+        """Return the host declaration name for DHCPD for this `interface`."""
+        interface_name = interface.name.replace(".", "-")
+        if interface.node_config_id is None:
+            return f"unknown-{interface.id}-{interface_name}"
```
MaikRe, line 1068: nit: missing test for this

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -579,116 +575,85 @@ async def get_active_interfaces_for_agent(
...
+                            interface_ids.add(parent.id)
+                            hostname = await self._make_interface_hostname(
+                                parent, svc
+                            )
+                            if parent.mac_address:
+                                host_reservations.append(
+                                    HostReservationData(
+                                        ip=str(sip.ip),
+                                        mac_address=parent.mac_address,
+                                        hostname=hostname,
```
MaikRe, line 625: https://ftp.ripe.net/mirrors/sites/ftp.isc.org/isc/kea/2.6.1/doc/html/arm/dhcp4-srv.html#multiple-reservations-for-the-same-ip

KEA seems to not allow this type of same ips within a given subnet.
Is there a setting somewhere that ensures KEA is compatible with this type of sharing?

MaikRe (commented):
Question/nit inline

src/maastemporalworker/workflow/dhcp.py
wyattrees: The code being removed here is dead code that was previously used for configuring the experimental internal DHCP server that we've since removed. The logic in this function simply needs to mimic how we configure the ISC dhcp server (i.e. what maasserver/dhcp.py::make_hosts_for_subnets does)

src/maastemporalworker/workflow/dhcp.py
wyattrees: per the link you shared, this is allowed if you set `"ip-reservations-unique": false` in the configuration

MaikRe: Is there any place in the code where this setting is set to false or is this something that happens external to MAAS?

wyattrees: This part of the configuration isn't implemented yet, I'll make a note to include this field

MaikRe (approved):
+1

PR #396, opened 2026-06-18
feat: start operation workflows after commit with OperationUUID

tanzin8r:
Add helper to create an `ACCEPTED` operation and register its Temporal workflow after commit. Workflow name is derived from op_type and parameters are persisted on the operation row and passed through to Temporal for reconciliation replay.

### DIFF
--- a/src/maasservicelayer/services/operations.py
+++ b/src/maasservicelayer/services/operations.py
@@ -76,7 +76,7 @@ class OperationsService(
         await self.set_current_task(operation_uuid, name)
         return task
 
-    async def start_operation(
+    async def create_accepted_operation(
         self,
         *,
         op_type: OperationType,
--- a/src/maasservicelayer/services/operations.py
+++ b/src/maasservicelayer/services/operations.py
@@ -1,7 +1,21 @@
 # Copyright 2026 Canonical Ltd.  This software is licensed under the
 # GNU Affero General Public License version 3 (see the file LICENSE).
 
-from maascommon.enums.operations import OperationStatus, OperationTaskStatus
+from typing import Any
+from uuid import uuid4
+
+from temporalio.common import (
+    SearchAttributeKey,
+    SearchAttributePair,
+    TypedSearchAttributes,
+)
+
+from maascommon.enums.operations import (
+    OperationStatus,
+    OperationTaskStatus,
+    OperationType,
+)
+from maascommon.workflows.operation import OPERATION_UUID_SEARCH_ATTRIBUTE
 from maasservicelayer.builders.operations import (
     OperationBuilder,
     OperationTaskBuilder,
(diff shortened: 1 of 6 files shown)


maas-lander:
- [x] Direct message
- [ ] ~maas

src/maasservicelayer/services/operations.py
```diff
@@ -57,6 +74,48 @@ async def start_task(
         await self.set_current_task(operation_uuid, name)
         return task
 
+    async def start_operation(
```
r00ta: `start_operation` would suggest that the operation is getting started, so it's a bit misleading

src/maasservicelayer/services/operations.py
```diff
@@ -57,6 +74,48 @@ async def start_task(
...
+        self,
+        *,
```
r00ta: why `*`?

r00ta (commented):
inline

src/maasservicelayer/services/operations.py
tanzin8r: Fair point since it only creates the `ACCEPTED` row and registers the workflow post-commit. How about `create_accepted_operation`?

src/maasservicelayer/services/operations.py
tanzin8r: Everything after `*` will be keyword only. I added it because this method has several optional params that are easy to mix up for example `parameters` is stored on the operation row, `workflow_parameter` goes to Temporal etc. This will force callers to name each one.

src/maasservicelayer/services/temporal.py
```diff
@@ -204,6 +197,7 @@ def register_workflow_call(
             parameter,
             workflow_id,
             wait,
+            search_attributes,
             args,
             kwargs,
         )
```
AloizioMacedo, line 192: The way `search_attributes` is being passed around is somewhat inconsistent. You are forcing it here as a keyword-only argument, but then you change the order of it with args when populating the key in the dict, and then in `post_commit` you inject it inside kwargs forcefully.

If it is going to be put into kwargs anyway, why not specify it as a kwarg in the function call when necessary? (You are already doing this in `create_accepted_operation`, for example.) I don't think you need to change any of the code in `temporal.py` if I am not mistaken.

tanzin8r: you're right. I was trying to make `search_attributes` a named parameter on `TemporalService`

src/maasservicelayer/services/operations.py
```diff
@@ -59,6 +78,50 @@ async def start_task(
...
+        operation = await self.create(
+            builder=OperationBuilder(
+                uuid=str(uuid4()),
+                op_type=op_type,
+                status=OperationStatus.ACCEPTED,
+                resource_id=resource_id,
+                resource_type=resource_type,
+                parameters=parameters,
+                user_id=user_id,
+                is_bulk=False,
```
r00ta, line 104: for bulk operation we have something else?

r00ta (commented):
question inline

src/maasservicelayer/services/operations.py
tanzin8r: Yes, this helper is for stand alone operations only. Bulk operation will create a parent row and then child rows will be created inside that workflow with `parent_id` set that is the next step.

AloizioMacedo (approved):

PR #397, opened 2026-06-19
docs: fix docs build for git shallow clones

alemar99:
The `sphinx_last_updated_by_git` will raise a warning when the git clone is shallow (e.g. when cloning with `git clone --depth 1`) causing the `make html` command to fail.


maas-lander:
- [ ] Direct message
- [ ] ~maas

alemar99:
This was also making our deb-based system-tests fail because they can't build the package.

troyanov (approved):

PR #398, opened 2026-06-19
docs: fix docs build for git shallow clones

alemar99:
The `sphinx_last_updated_by_git` will raise a warning when the git clone is shallow (e.g. when cloning with `git clone --depth 1`) causing the `make html` command to fail.

(cherry picked from commit 60e182d44a64c5ddef909eede7b95aa7b1387d22)


maas-lander:
- [ ] Direct message
- [ ] ~maas

PR #400, opened 2026-06-19
refactor: change roles to groups in /users endpoints

nehjoshi:
- When fetching the users list, we no longer require the `roles` column as it is irrelevant 
- Instead, we should fetch the groups the user belongs to
- Same with the `POST /users` and `/PUT /users/:id` endpoints; we should allow users to be added to/removed from groups in those forms


maas-lander:
- [x] Direct message
- [ ] ~maas

src/maasapiserver/v3/api/public/models/requests/users.py
```diff
@@ -63,13 +62,17 @@ def check_email(cls, v: str | None) -> str | None:
...
+        description="The IDs of the groups the user will be a member of.",
+    )
 
     def to_builder(self) -> UserBuilder:
         hashed_password = UserBuilder.hash_password(self.password)
         return UserBuilder(
             username=self.username,
             password=hashed_password,
-            is_superuser=self.is_superuser,
+            is_superuser=False,
```
nehjoshi, line 75: This is irrelevant, but I've left it in as the builder still expects `is_superuser`

src/maasapiserver/v3/api/public/handlers/users.py
```diff
@@ -324,9 +352,37 @@ async def update_user(
             user_id, user_request.to_builder()
         )
 
+        current_groups = (
+            await services.users.get_groups_for_users([user.id])
+        ).for_user(user.id)
+        current_group_ids = {group.id for group in current_groups}
+        desired_group_ids = dict.fromkeys(user_request.groups)
```
alemar99:
```suggestion
        desired_group_ids = set(user_request.groups)
```

src/maasapiserver/v3/api/public/handlers/users.py
```diff
@@ -324,9 +352,37 @@ async def update_user(
...
+
+        for group_id in desired_group_ids:
+            if group_id in current_group_ids:
+                continue
```
alemar99:
```suggestion
        for group_id in desired_group_ids - current_group_ids:
```

src/maasapiserver/v3/api/public/handlers/users.py
```diff
@@ -324,9 +352,37 @@ async def update_user(
...
+                            type=INVALID_ARGUMENT_VIOLATION_TYPE,
+                            message=f"Group with id '{group_id}' does not exist.",
+                        )
+                    ]
+                ) from err
+
+        for group_id in current_group_ids - desired_group_ids.keys():
+            await services.usergroups.remove_user_from_group(group_id, user.id)
+
+        groups_by_user = await services.users.get_groups_for_users([user.id])
```
alemar99, line 384: Is this necessary?

src/maasapiserver/v3/api/public/handlers/users.py
```diff
@@ -324,9 +352,37 @@ async def update_user(
...
+                raise NotFoundException(
+                    details=[
+                        BaseExceptionDetail(
+                            type=INVALID_ARGUMENT_VIOLATION_TYPE,
+                            message=f"Group with id '{group_id}' does not exist.",
+                        )
+                    ]
+                ) from err
+
+        for group_id in current_group_ids - desired_group_ids.keys():
```
alemar99:
```suggestion
        for group_id in current_group_ids - desired_group_ids:
```

alemar99 (commented):
Some suggestions inline

src/maasapiserver/v3/api/public/handlers/users.py
nehjoshi: Yes; the other approach would be to compute the new groups locally, but because we only have access to `group_id` here, we would need to create separate DB calls to get the name and description for each group, which is a lot more cumbersome.

nehjoshi:
Committed the suggestions from review comments ->

alemar99 (dismissed):
+1

AloizioMacedo:
There is a method called `list` in `users.py` which is messing up some things, including the parsing that is used in the docs but also the typing itself in `users.py` after its definition:

image

As far as I understand, this is due to how Python builds the class and the fact that while building, it populates a local namespace with the method names.

I think the easiest way to avoid this issue is to simply use `List` from typing here. Or add `from __future__ import annotations` in the top, but I've seen this last one have issues with pydantic in the past.

It should be okay to use built-in names as method names, but there are those little quirks around it. Perhaps we should avoid it in the future, particularly if they relate to type hints.

nehjoshi:
Thanks @AloizioMacedo , I'll change it to `List`. I need to add another commit anyway so I will include the change with that commit

nehjoshi:
In the most recent commit, I've modified the `/users/me` endpoint to return the user's entitlements instead of `is_superuser`. This will be used by the UI to show/hide certain pages and actions. Files of interest:
- `src/maasapiserver/v3/api/public/handlers/users.py`
- `src/maasapiserver/v3/api/public/models/responses/users.py`
- `src/maasservicelayer/services/openfga_tuples.py`

src/maasapiserver/v3/api/public/handlers/users.py
```diff
@@ -117,8 +121,25 @@ async def get_user_info(
...
+            for t in entitlement_tuples
+        }
+        entitlements = [
+            EntitlementResponse.from_model(t) for t in unique.values()
+        ]
         return UserInfoResponse(
-            id=user.id, username=user.username, is_superuser=user.is_superuser
+            id=user.id,
+            username=user.username,
+            entitlements=entitlements,
```
alemar99: Here we should pass the entitlements and build the `EntitlementResponse` in the `UserInfoResponse`, like you did for groups in `UserResponse`

src/maasservicelayer/services/openfga_tuples.py
```diff
@@ -275,6 +275,19 @@ async def list_entitlements(
...
         return await self.get_many(query)
 
+    async def list_entitlements_for_groups(
+        self,
+        group_ids: list[int],
+    ) -> list[OpenFGATuple]:
+        if not group_ids:
+            return []
+        query = QuerySpec(
+            where=OpenFGATuplesClauseFactory.with_users(
```
alemar99, line 285: Not related to this PR: `with_users` is a bit misleading when we then use it with a group. We may think of a better naming in the future (or just leave as it is)

alemar99 (commented):
looks good, one nit inline

src/maasservicelayer/services/openfga_tuples.py
nehjoshi: Agreed. I will take note of this and follow it up in a separate PR

src/maasapiserver/v3/api/public/handlers/users.py
nehjoshi: Good catch!

alemar99 (approved):
LGTM

PR #402, opened 2026-06-23
fix: re-fetch the manifest when updating the url of a boot source

alemar99:


maas-lander:
- [ ] Direct message
- [ ] ~maas

src/maasserver/api/tests/test_boot_sources.py
```diff
@@ -142,7 +144,23 @@ def test_PUT_calls_workflow_if_url_changed(self):
         self.assertEqual(http.client.OK, response.status_code)
         boot_source = reload_object(boot_source)
         self.assertEqual(boot_source.url, new_values["url"])
-        mock_post_commit_do.assert_called_once()
+
+        calls = mock_post_commit_do.call_args_list
+        # Ugly, but since the method is bound to the boot source object returned
```
nehjoshi, line 149: What if we do:
```py
calls = [c.args[2] for c in mock_post_commit_do.call_args_list]

self.assertEqual(len(calls), 2)

self.assertEqual(
calls[0].__func__,
    BootSource.verify_selections_after_url_update,
)
self.assertEqual(
calls[1].__func__,
    BootSource.refetch_manifest,
)
```
?

nehjoshi (commented):
Looks good, just one suggestion ->

src/maasserver/api/tests/test_boot_sources.py
alemar99: Doing so we'll lose the assert to `reactor.callLater`, no?

nehjoshi: Ah yes, my bad. In that case, I don't see a better way to do this than to index each property

nehjoshi (approved):
LGTM

PR #403, opened 2026-06-23
refactor(bootresources): better exception handling for boot resources download

alemar99:
When downloading images, 4xx and 5xx errors were treated as retryable errors causing the activity to retry indefinitely.


maas-lander:
- [ ] Direct message
- [ ] ~maas

nehjoshi (approved):
+1

PR #409, opened 2026-06-24
fix(bootresources): match all lower priorities selections when deleting resources

alemar99:
We were using `get_one`, but in fact there could be more than one selection matching the query. This resulted in an exception which caused the activity to retry indefinitely.

### DIFF
--- a/src/tests/maastemporalworker/workflow/test_bootresource.py
+++ b/src/tests/maastemporalworker/workflow/test_bootresource.py
@@ -824,7 +824,7 @@ class TestGetFilesToDownloadForSelectionActivity:
         services_mock.boot_source_selections.get_by_id.return_value = (
             boot_source_selection
         )
-        services_mock.boot_source_selections.get_one.return_value = (
+        services_mock.boot_source_selections.get_many.return_value = [
             BootSourceSelection(
                 id=2,
                 os="ubuntu",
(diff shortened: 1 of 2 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

src/tests/maastemporalworker/workflow/test_bootresource.py
```diff
@@ -824,7 +824,7 @@ async def test_calls_service_layer(
         services_mock.boot_source_selections.get_by_id.return_value = (
             boot_source_selection
         )
-        services_mock.boot_source_selections.get_one.return_value = (
+        services_mock.boot_source_selections.get_many.return_value = [
             BootSourceSelection(
```
nehjoshi, line 828: Do you think we should maybe add one more `BootSourceSelection` in this array to test only the correct ones are filtered? (Since we are now using `get_many` instead of `get_one`)

nehjoshi (commented):
Looks good, just one question

src/tests/maastemporalworker/workflow/test_bootresource.py
alemar99: Given the fact that this is mock, this will always return 2 selections and the filtering would not be tested. Or maybe I misunderstood your question.
What we could do is returning 2 selections and assert that for every of them the `services_mock.boot_resources.delete_many` is called.

nehjoshi: Yep, the first point is what I was talking about, thanks!

nehjoshi (approved):
LGTM

PR #410, opened 2026-06-24
fix(bootresources): stop other fetch-manifest workflows when updating all the manifests

alemar99:
If the fetch-manifest workflows try to update simultaneously the manifest for the same boot source, serialization errors will happen.


maas-lander:
- [ ] Direct message
- [ ] ~maas

nehjoshi (approved):
LGTM

PR #413, opened 2026-06-24
docs: patch docs cryptography dependency

AloizioMacedo:
This allows us to bypass issues when attempting to inadvertently compile `cryptography` on s390x and perhaps other architectures, since the wheels for them are not readily available and thus running `pip install` will try to force the build of an underlying pyo3 rust-python project.

### DIFF
--- a/docs/_scripts/get_api_spec.py
+++ b/docs/_scripts/get_api_spec.py
@@ -8,6 +8,7 @@ from importlib.abc import Loader
 from importlib.machinery import ModuleSpec
 import os
 import sys
+from types import ModuleType
 from typing import Any
 from unittest.mock import MagicMock
 
@@ -17,6 +18,39 @@ try:
     import importlib.metadata as _ilm
 except ImportError:
     _ilm = None
+
+
+def _patch_recursively(root_mock: MagicMock, module_name: str):
+    # The trickery that follows is intended to prevent two things:
+    # 1 - A bunch of assignments to sys.modules
+    # 2 - Maintenance here in case we simply import a different submodule
+    #     of the patched module.
+    class PatchedFinder:
+        def find_spec(self, fullname, path, target=None):
+            if fullname == module_name or fullname.startswith(
+                f"{module_name}."
+            ):
+                return ModuleSpec(fullname, PatchedLoader())
+            return None
+
+    class PatchedLoader(Loader):
+        def create_module(self, spec):
+            module = ModuleType(spec.name)
+
+            parts = spec.name.split(".")[1:]
+            mock_obj = root_mock
+            for part in parts:
+                mock_obj = getattr(mock_obj, part)
+
+            module.__path__ = []  # Triggers Python to treat it as a package
+            module.__getattr__ = lambda name: getattr(mock_obj, name)
+
+            return module
+
+        def exec_module(self, module):
+            pass
+
+    sys.meta_path.insert(0, PatchedFinder())
 
 
 def _patch_maas_metadata():
@@ -92,6 +126,48 @@ def _patch_tftp():
     sys.modules["tftp.backend"] = tftp.backend
 
 
+def _patch_authlib():
+    """Patch authlib so imports succeed."""
+    authlib = MagicMock()
+    _patch_recursively(authlib, "authlib")
+
+
+def _patch_cryptography():
+    """Patch cryptography so imports succeed."""
+    cryptography = MagicMock()
+    _patch_recursively(cryptography, "cryptography")
+
+
+def _patch_joserfc():
+    """Patch cryptography so imports succeed."""
+    joserfc = MagicMock()
+    _patch_recursively(joserfc, "joserfc")
+
+
+def _patch_pylxd():
+    """Patch cryptography so imports succeed."""
+    pylxd = MagicMock()
+    pylxd.client = MagicMock()
+    pylxd.exceptions = MagicMock()
+    sys.modules["pylxd"] = pylxd
+    sys.modules["pylxd.client"] = pylxd.client
+    sys.modules["pylxd.exceptions"] = pylxd.exceptions
+
+
+def _patch_openssl():
+    """Patch OpenSSL so imports succeed.
+
+    In this case, we do not avoid openssl altogether. Rather,
+    only some paths that interact with cryptography through twisted.
+    """
+    openssl = MagicMock()
+    openssl._util = MagicMock()
+    openssl.SSL = MagicMock()
+    sys.modules["OpenSSL"] = openssl
+    sys.modules["OpenSSL.SSL"] = openssl.SSL
+    sys.modules["OpenSSL._util"] = openssl._util
+
+
 def _patch_temporalio():
     """Patch temporalio module so imports succeed without the package installed."""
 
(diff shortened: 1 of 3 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

alemar99:
I get the following error when trying to build the docs on a s390x (emulated) env:
```
root@d9285aa521f5:/maas/docs# make html
if git status > /dev/null; then \
        git clean -fx "_build"; \
fi
rm -rf .sphinx/.doctrees
... setting up virtualenv
python3 -m venv .sphinx/venv || { echo "You must install python3-venv and python3-dev before you can build the documentation."; exit 1; }
. .sphinx/venv/bin/activate; pip install pip-tools
Collecting pip-tools
  Downloading pip_tools-7.5.3-py3-none-any.whl.metadata (27 kB)
Collecting build>=1.0.0 (from pip-tools)
  Downloading build-1.5.0-py3-none-any.whl.metadata (5.7 kB)
Collecting click>=8 (from pip-tools)
  Downloading click-8.4.2-py3-none-any.whl.metadata (2.6 kB)
Requirement already satisfied: pip>=22.2 in ./.sphinx/venv/lib/python3.12/site-packages (from pip-tools) (24.0)
Collecting pyproject_hooks (from pip-tools)
  Downloading pyproject_hooks-1.2.0-py3-none-any.whl.metadata (1.3 kB)
Collecting setuptools (from pip-tools)
  Downloading setuptools-82.0.1-py3-none-any.whl.metadata (6.5 kB)
Collecting wheel (from pip-tools)
  Downloading wheel-0.47.0-py3-none-any.whl.metadata (2.3 kB)
Collecting packaging>=24.0 (from build>=1.0.0->pip-tools)
  Downloading packaging-26.2-py3-none-any.whl.metadata (3.5 kB)
Downloading pip_tools-7.5.3-py3-none-any.whl (71 kB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 71.4/71.4 kB 940.1 kB/s eta 0:00:00
Downloading build-1.5.0-py3-none-any.whl (26 kB)
Downloading click-8.4.2-py3-none-any.whl (119 kB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 119.2/119.2 kB 934.6 kB/s eta 0:00:00
Downloading pyproject_hooks-1.2.0-py3-none-any.whl (10 kB)
Downloading setuptools-82.0.1-py3-none-any.whl (1.0 MB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.0/1.0 MB 841.9 kB/s eta 0:00:00
Downloading wheel-0.47.0-py3-none-any.whl (32 kB)
Downloading packaging-26.2-py3-none-any.whl (100 kB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100.2/100.2 kB 3.3 MB/s eta 0:00:00
Installing collected packages: setuptools, pyproject_hooks, packaging, click, wheel, build, pip-tools
Successfully installed build-1.5.0 click-8.4.2 packaging-26.2 pip-tools-7.5.3 pyproject_hooks-1.2.0 setuptools-82.0.1 wheel-0.47.0
. .sphinx/venv/bin/activate; pip-compile requirements.txt --output-file=compiled-requirements.txt
    error: subprocess-exited-with-error

    × Getting requirements to build wheel did not run successfully.
    │ exit code: 1
    ╰─> [3 lines of output]
        Building lxml version 6.0.2.
        Building without Cython.
        Error: Please make sure the libxml2 and libxslt development packages are installed.
        [end of output]
```

alemar99:
If I install those suggested packages, I get another error:
```
. .sphinx/venv/bin/activate; pip-compile requirements.txt --output-file=compiled-requirements.txt
    error: subprocess-exited-with-error

    × python setup.py egg_info did not run successfully.
    │ exit code: 1
    ╰─> [36 lines of output]
        /maas/docs/.sphinx/venv/lib/python3.12/site-packages/setuptools/dist.py:765: SetuptoolsDeprecationWarning: License classifiers are deprecated.
        !!

                ********************************************************************************
                Please consider removing the following classifiers in favor of a SPDX license expression:

                License :: OSI Approved :: GNU Library or Lesser General Public License (LGPL)

                See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
                ********************************************************************************

        !!
          self._finalize_license_expression()
        running egg_info
        creating /tmp/pip-pip-egg-info-sg6zdso5/psycopg2_binary.egg-info
        writing /tmp/pip-pip-egg-info-sg6zdso5/psycopg2_binary.egg-info/PKG-INFO
        writing dependency_links to /tmp/pip-pip-egg-info-sg6zdso5/psycopg2_binary.egg-info/dependency_links.txt
        writing top-level names to /tmp/pip-pip-egg-info-sg6zdso5/psycopg2_binary.egg-info/top_level.txt
        writing manifest file '/tmp/pip-pip-egg-info-sg6zdso5/psycopg2_binary.egg-info/SOURCES.txt'

        Error: pg_config executable not found.

        pg_config is required to build psycopg2 from source.  Please add the directory
        containing pg_config to the $PATH or specify the full executable path with the
        option:

            python setup.py build_ext --pg-config /path/to/pg_config build ...

        or with the pg_config option in 'setup.cfg'.

        If you prefer to avoid building psycopg2 from source, please install the PyPI
        'psycopg2-binary' package instead.

        For further information please check the 'doc/src/install.rst' file (also at
        <https://www.psycopg.org/docs/install.html>).

        [end of output]

    note: This error originates from a subprocess, and is likely not a problem with pip.
```

alemar99:
Maybe we can continue doing it in the same way as of now and we just patch authlib which requires cryptography? I'm testing it now

AloizioMacedo:
>Maybe we can continue doing it in the same way as of now and we just patch authlib which requires cryptography? I'm testing it now

It is not only authlib which requires cryptography : /, that is part of the issue:

```bash
Using Python 3.12.3 environment at: .sphinx/venv
Name: cryptography
Version: 46.0.7
Location: /home/aloizio.macedo@canonical.com/Canonical/maas/docs/.sphinx/venv/lib/python3.12/site-packages
Requires: cffi
Required-by: authlib, joserfc, paramiko, pylxd, pyopenssl
```

It seems that the problem is related directly to pip-compile (which is a little surprising to be honest). But we could just commit the compiled-requirements.txt directly, it would even save some time for the build, and make it clear that it is an output of requirements.txt. 

On the other hand, it is not that many required-bys. I will see if it is feasible to patch them. If it is, we can even revert having the compiled-reqs. If it is not, I will commit it.

Update: It was fine patching them all, pushed changes that do so.

alemar99:
I've tried the latest changes and I got the same errors above (regarding `lxml` and `psycopg2`).
I was thinking on this problem and the solution could be the following:
- we install the python3 system packages for the packages that don't ship wheels for s390x
- we create the venv with `--system-site-packages`
- we keep those packages in the requirements.txt and we `pip install` them WITHOUT the `--upgrade` flag (I don't know if it was there for some specific reasons).

Then, for snaps, we have to include these packages in the `offline-docs` part under the `build-packages` section.
For debs, instead, we don't have to do anything since we pre-built and bundle them.
For local development, we usually have most of the packages already installed since we are developing MAAS, so that would not change much.

We could also add a variable in the makefile to for creating the venv with the `--system-site-packages` or not. We would then use it only on the snapcraft.

This would solve (hopefully) all the issues, what do you think?

AloizioMacedo:
@alemar99 ReadTheDocs does not really support --system-site-packages. I've done some research around this, and basically it seems that they deprecated some specific features of their configuration around this (see https://github.com/WMD-group/SMACT/issues/155) and if you try to override things by changing the `create_environment` setup, it still has some internal wrapping that disregards your installation wrt the virtual env. 

Even if it did support, it does not support custom PPAs, which we would need (or patch around).

In any case, I gave some tries to the idea and you can see some tests run failures in our RTD dashboard.

So it would be a lot of stitching to try to make it work there. The fact is that our build process and environment is very bespoke, and RTD does not support it, at least not cleanly. 

---

Ideally, I would also prefer that the method we choose would homogeneously apply to all cases we have (snap and deb for all archs, and RTD) without a lot of case-by-case solutions, even if it, itself, were a little bit ugly (i.e., has a lot of patches etc). There is no way to definitely avoid patching anyway, since some of the dependencies of the docs themselves might need patching (for instance, some of the ones related to images required special apt packages).

I will see what I can do with lxml and psycopg2. I think and hope that it is just a matter of some system dependencies that we are not including in the snapcraft, and should be relatively straightforward to infer from the maas part of the snap. (OBS: The real solution around all this imo is getting rid of the dedicated API and CLI docs pages, but we can't do this right now.)

alemar99:
Ah, didn't know they don't support `--system-site-packages`...

Btw, I see you added some packages to `stage-packages` in snapcraft.yaml. Since those are only needed for s390x, do you think it will be worth it to use `to s390x:` directive? I saw that in the [snapcraft docs](https://documentation.ubuntu.com/snapcraft/9.0/reference/advanced-grammar/) but to be honest I never used it

AloizioMacedo:
> Ah, didn't know they don't support `--system-site-packages`...
> 
> Btw, I see you added some packages to `stage-packages` in snapcraft.yaml. Since those are only needed for s390x, do you think it will be worth it to use `to s390x:` directive? I saw that in the [snapcraft docs](https://documentation.ubuntu.com/snapcraft/9.0/reference/advanced-grammar/) but to be honest I never used it

Most are a subset of the system dependencies of the `maas` part, so I think it is worth leaving there. The ones that are not (e.g. libcairo2) are needed in any architecture.

alemar99 (approved):
LGTM, great!

PR #416, opened 2026-06-25
fix(bootresources): don't block the API handler when importing boot resources

alemar99:
When you do `maas <profile> boot-resources import` the API response is not returned until the fetch manifest workflow terminates. To avoid this, we schedule this to be called in a thread and return immediately.

### DIFF
--- a/src/maasserver/bootresources.py
+++ b/src/maasserver/bootresources.py
@@ -18,6 +18,7 @@ from django.db.models import F
 from temporalio.client import WorkflowExecutionStatus
 from temporalio.common import WorkflowIDReusePolicy
 from twisted.application.internet import TimerService
+from twisted.internet import reactor
 from twisted.internet.defer import inlineCallbacks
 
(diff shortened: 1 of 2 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

nehjoshi (commented):
Looks good. I noticed we don't have a test covering the `import_resources` func. Would something minimal like this be appropriate?
```py
class TestImportResources(MAASServerTestCase):
    def test_schedules_start_on_reactor_and_returns_immediately(self):
        mock_reactor = self.patch(bootresources, "reactor")

        result = bootresources.import_resources()

        self.assertIsNone(result)
        mock_reactor.callFromThread.assert_called_once()
        [scheduled], _ = mock_reactor.callFromThread.call_args
        self.assertTrue(callable(scheduled))
```

nehjoshi (approved):
LGTM!

PR #417, opened 2026-06-26
fix: add nosniff security header to nginx configuration

r00ta:
Disable content-type sniffing on some browsers


maas-lander:
- [ ] Direct message
- [ ] ~maas

alemar99 (approved):
+1

PR #419, opened 2026-06-26
docs: patch docs cryptography dependency

alemar99:
This allows us to bypass issues when attempting to inadvertently compile `cryptography` on s390x and perhaps other architectures, since the wheels for them are not readily available and thus running `pip install` will try to force the build of an underlying pyo3 rust-python project.

(cherry picked from commit 8e10cded9c521647190fe11e7b544eda411bdd00)


maas-lander:
- [ ] Direct message
- [ ] ~maas

PR #425, opened 2026-06-29
feat: add client credentials flow for OIDC CLI adaption

nehjoshi:
This PR introduces the [client credentials flow](https://auth0.com/docs/authenticate/login/oidc-conformant-authentication/oidc-adoption-client-credentials-flow) in order to solve the problem of OIDC users unable to sign in via the CLI. The approach is the following:
1. An OIDC user tries to sign in via the CLI/APIv2
2. v2 forwards the request to the service layer in v3
3. If the user is indeed an OIDC user, we get the corresponding adapter for the enabled provider
4. MAAS makes a machine-to-machine request to the provider for an access token
5. MAAS queries the provider's `/users` endpoint using the access token to check if the user actually exists and is enabled on the provider side
6. If a generic provider is enabled (i.e. not one of EntraID, Auth0 or Keycloak), it is the admin's responsibility to ensure the user exists and is enabled. MAAS will not perform this check in this case.

To verify this, a QA with Auth0 as a provider might be helpful:
1. See steps for setting up Auth0 and adding it to MAAS [here](https://pastebin.canonical.com/p/3p3P4Y9Q3P/)
2. Ensure you have pulled this branch and have done `make snap-tree-sync && snap restart maas`
3. First, sign in via the UI as an OIDC user. This will ensure a new user is created in the DB. If you've already done this step before and have an existing OIDC user, you can go to step 4.
4. Generate an API key for the OIDC user.
5. Sign in via the CLI as the OIDC user:
```sh
maas login -k \
  '<YOUR_EMAIL>' \
  https://localhost:5443/MAAS \
  '<API_KEY>'
```
6. Ensure you are signed in and can use the CLI normally

## Notes
- [This](https://docs.google.com/document/d/1AVFwdXHIJDsFLr7A8BXOB0UzVh_yUrfn_00LEsAExg8/edit?usp=sharing) document containing notes on provider API formats could be useful

### DIFF
--- a/src/maasapiserver/v3/api/public/models/requests/external_auth.py
+++ b/src/maasapiserver/v3/api/public/models/requests/external_auth.py
@@ -46,7 +46,6 @@ class OAuthProviderRequest(BaseModel):
         description="The type of access tokens issued by the OIDC provider (e.g., JWT or opaque).",
     )
     vendor: OAuthVendorChoices = Field(
-        default=OAuthVendorChoices.GENERIC,
         description="The OIDC provider vendor. Select from predefined vendors or use 'Generic' for other providers.",
     )
     scopes: str = Field(
--- a/src/maasapiserver/v3/api/public/models/requests/external_auth.py
+++ b/src/maasapiserver/v3/api/public/models/requests/external_auth.py
@@ -8,12 +8,22 @@ from pydantic import BaseModel, Field, field_validator, ValidationInfo
 
 from maasservicelayer.builders.external_auth import OAuthProviderBuilder
 from maasservicelayer.exceptions.catalog import ValidationException
-from maasservicelayer.models.external_auth import AccessTokenType
+from maasservicelayer.models.external_auth import (
+    AccessTokenType,
+    ProviderVendorType,
+)
 
 
 class OAuthTokenTypeChoices(StrEnum):
     JWT = "JWT"
     OPAQUE = "Opaque"
+
+
+class OAuthVendorChoices(StrEnum):
+    GENERIC = "Generic"
+    ENTRAID = "EntraID"
+    AUTH0 = "Auth0"
+    KEYCLOAK = "Keycloak"
 
 
 class OAuthProviderRequest(BaseModel):
@@ -34,6 +44,10 @@ class OAuthProviderRequest(BaseModel):
     )
     token_type: OAuthTokenTypeChoices = Field(
         description="The type of access tokens issued by the OIDC provider (e.g., JWT or opaque).",
+    )
+    vendor: OAuthVendorChoices = Field(
+        default=OAuthVendorChoices.GENERIC,
+        description="The OIDC provider vendor. Select from predefined vendors or use 'Generic' for other providers.",
     )
     scopes: str = Field(
         description="A space-separated list of OIDC scopes defining the information requested from the provider.",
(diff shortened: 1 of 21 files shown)


maas-lander:
- [x] Direct message
- [ ] ~maas

src/maasservicelayer/auth/oidc_adapters.py
```diff
@@ -0,0 +1,185 @@
...
+
+    @abstractmethod
+    def _token_request_body(self) -> dict[str, str]:
+        """Provider-specific body for the client credentials token request."""
+
+    async def _token_request(
+        self, url: str, data: dict[str, str]
+    ) -> dict[str, Any]:
+        response = await self._http.post(url, data=data)
+        response.raise_for_status()
```
alemar99: I think we should handle exceptions here instead of making them bubble up. We should make the user aware that the provider is probably misconfigured and they have to take action.

alemar99 (changes requested):
Looks good, see inline.

Also another question: is there a reason why we are defaulting to generic vendor for the OIDC provider?

src/maasservicelayer/auth/oidc_adapters.py
nehjoshi: Good idea, this has been added

nehjoshi:
> Also another question: is there a reason why we are defaulting to generic vendor for the OIDC provider?

Because there is no way to deduce the provider vendor, we have to default to a Generic provider (for existing providers that don't have vendors). This was mainly to ensure existing latest/edge installations (like, the demo instance, for example) don't break when we introduce this new column. If it's deemed unnecessary, I am happy to remove the default

alemar99:
> > Also another question: is there a reason why we are defaulting to generic vendor for the OIDC provider?
> 
> Because there is no way to deduce the provider vendor, we have to default to a Generic provider (for existing providers that don't have vendors). This was mainly to ensure existing latest/edge installations (like, the demo instance, for example) don't break when we introduce this new column. If it's deemed unnecessary, I am happy to remove the default

We shouldn't care about latest/edge installations in general, they are supposed to break and we give no guarantees on them. I'd make it a required parameter on the API side and remove the default in the db, so the user has to specify which vendor to use, to avoid mistakes like adding a supported provider but ending up with setting it as Generic.

nehjoshi:
Thanks for the detailed explanation; I've removed the default value

src/maasapiserver/v3/api/public/models/requests/external_auth.py
```diff
@@ -35,6 +45,10 @@ class OAuthProviderRequest(BaseModel):
     token_type: OAuthTokenTypeChoices = Field(
         description="The type of access tokens issued by the OIDC provider (e.g., JWT or opaque).",
     )
+    vendor: OAuthVendorChoices = Field(
+        default=OAuthVendorChoices.GENERIC,
```
alemar99: IMO we shouldn't default to anything

src/maasserver/api/oidc_providers.py
```diff
@@ -280,12 +308,14 @@ def create(self, request):
         client_secret = request.POST.get("client_secret", None)
         enabled = request.POST.get("enabled", "false").lower() == "true"
         token_type = request.POST.get("token_type", "JWT")
+        vendor = request.POST.get("vendor", "Generic")
```
alemar99: same as above, we should make this required.
Also I'm wondering, why in v3 all the fields are required whereas in v2 many of them are not?

nehjoshi: Likely as a result of directly copying the update() method. This is fixed.

alemar99 (dismissed):
LGTM. I don't know if you have to update some tests after changing fields to be required, approving meanwhile

alemar99 (dismissed):

alemar99 (approved):

PR #426, opened 2026-06-29
feat: reconcile operations when Temporal workflow has terminated

tanzin8r:
Extend the scheduled reconcile-operations workflow to fix RUNNING and CANCELLING operations when the linked Temporal execution has already finished or been terminated externally. I also refactored so both reconciliation phases run as batch activities.

### DIFF
--- a/src/maastemporalworker/workflow/operation.py
+++ b/src/maastemporalworker/workflow/operation.py
@@ -142,32 +142,32 @@ class OperationActivity(ActivityBase):
             operations = (
                 await services.operations.list_in_progress_operations()
             )
-        for operation in operations:
-            handle = self.temporal_client.get_workflow_handle(operation.uuid)
-            try:
-                description = await handle.describe()
-            except RPCError as e:
-                if e.status == RPCStatusCode.NOT_FOUND:
-                    if operation.status == OperationStatus.CANCELLING:
-                        status = OperationStatus.CANCELLED
-                        error = None
-                    else:
-                        status = OperationStatus.FAILED
-                        error = "Workflow execution not found in Temporal."
-                    async with self.start_transaction() as services:
+            for operation in operations:
+                handle = self.temporal_client.get_workflow_handle(
+                    operation.uuid
+                )
+                try:
+                    description = await handle.describe()
+                except RPCError as e:
+                    if e.status == RPCStatusCode.NOT_FOUND:
+                        if operation.status == OperationStatus.CANCELLING:
+                            status = OperationStatus.CANCELLED
+                            error = None
+                        else:
+                            status = OperationStatus.FAILED
+                            error = "Workflow execution not found in Temporal."
                         await services.operations.update_status(
                             operation_uuid=operation.uuid,
                             status=status,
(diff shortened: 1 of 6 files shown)


maas-lander:
- [x] Direct message
- [ ] ~maas

src/maastemporalworker/workflow/operation.py
```diff
@@ -99,54 +98,71 @@ async def update_current_task(self, param: UpdateCurrentTaskParam) -> None:
...
+            operations = (
+                await services.operations.list_in_progress_operations()
             )
+        for operation in operations:
+            handle = self.temporal_client.get_workflow_handle(operation.uuid)
+            try:
+                description = await handle.describe()
+            except RPCError as e:
+                if e.status == RPCStatusCode.NOT_FOUND:
+                    continue
```
wyattrees, line 170: if the workflow doesn't exist, shouldn't we update the status of the operation in the database?

src/maastemporalworker/workflow/operation.py
```diff
@@ -99,54 +98,71 @@ async def update_current_task(self, param: UpdateCurrentTaskParam) -> None:
...
+                raise
+            temporal_status = description.status
+            if temporal_status is None:
+                raise ValueError(
```
wyattrees: I don't think `ValueError` is correct here, since `temporal_status` is provided by Temporal and not the function arguments. Also, what does this actually mean if there is no status? Do we need to update the database?

wyattrees (commented):
couple of questions in-line

src/maastemporalworker/workflow/operation.py
tanzin8r: you are right. we should also update the db even if corresponding workflow doesn't exist. I am thinking we can update it as `cancelled` or `failed` depending on the db status. I'll make the changes.

src/maastemporalworker/workflow/operation.py
tanzin8r: technically it should never happen in practice. I added it as an edge case since the SDK types it as optional, but Temporal should always return a real status. Also, we’d skip the DB update since there would be no terminal status to map.

src/maastemporalworker/workflow/operation.py
```diff
@@ -99,54 +98,80 @@ async def update_current_task(self, param: UpdateCurrentTaskParam) -> None:
...
+                description = await handle.describe()
+            except RPCError as e:
+                if e.status == RPCStatusCode.NOT_FOUND:
+                    if operation.status == OperationStatus.CANCELLING:
+                        status = OperationStatus.CANCELLED
+                        error = None
+                    else:
+                        status = OperationStatus.FAILED
+                        error = "Workflow execution not found in Temporal."
+                    async with self.start_transaction() as services:
```
wyattrees: let's make everything in this activity use the same database transaction (including down below at the end).

wyattrees (changes requested):

wyattrees (approved):
lgtm

PR #427, opened 2026-06-30
fix: enable maas-agent metrics endpoint by default

skatsaounis:
The agent daemon rewrite gated `/metrics` behind `observability.metrics.enabled`, but the rackd-managed `agent.yaml` writer never emitted that flag, so `:5248/metrics/agent` returned 404. Emit `metrics.enabled=true` from RackAgent to restore the 3.7 behavior of the agent always serving `/metrics`.

This is a regression introduced after 3.7. On 3.7 `setupMetrics()` ran `mux.Handle("/metrics", ...)` unconditionally and was called unconditionally at startup (`Metrics.Enabled` was an unused placeholder): https://github.com/canonical/maas/blob/5aa0e0c1d9d4b88b812c8986795acd57669e5f55/src/maasagent/cmd/maas-agent/main.go#L271-L294 https://github.com/canonical/maas/blob/5aa0e0c1d9d4b88b812c8986795acd57669e5f55/src/maasagent/cmd/maas-agent/main.go#L413

The daemon rewrite changed this:

- commit 661491c7e defaulted observability.metrics.enabled to false, and
- commit 06652ce74 made setupMetrics gate registration on that flag, so the endpoint now 404s by default.
 
This fix targets the rackd-managed agent, whose agent.yaml is generated by RackAgent and never included the metrics flag.

### DIFF
--- a/src/provisioningserver/agent/config.py
+++ b/src/provisioningserver/agent/config.py
@@ -43,6 +43,11 @@ class AgentLogConfig:
 
 
 @dataclass(frozen=True)
+class AgentMetricsConfig:
+    enabled: bool
+
+
+@dataclass(frozen=True)
 class AgentServicesConfig:
     http_proxy: AgentHTTPProxyConfig
 
@@ -50,6 +55,7 @@ class AgentServicesConfig:
 @dataclass(frozen=True)
 class AgentObservabilityConfig:
     logging: AgentLogConfig
+    metrics: AgentMetricsConfig
 
 
 @dataclass(frozen=True)
--- a/src/provisioningserver/rackdservices/external.py
+++ b/src/provisioningserver/rackdservices/external.py
@@ -381,6 +381,7 @@ class RackAgent(RackOnlyExternalService):
                 logging=agent_config.AgentLogConfig(
                     "debug" if debug_enabled else "info",
                 ),
+                metrics=agent_config.AgentMetricsConfig(enabled=True),
             ),
             tls=agent_config.AgentTLSConfig(
                 key_file=key_file, cert_file=cert_file, ca_file=ca_file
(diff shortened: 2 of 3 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

skatsaounis:
# QA

## snap from latest master

```bash
ubuntu@juju-760bdc-0:~$ sudo cat /var/snap/maas/common/etc/maas/agent.yaml
# WARNING: MAAS will overwrite any changes made here.
...
observability:
  logging:
    level: info
services:
...

ubuntu@juju-760bdc-0:~$ curl  http://10.226.71.85:5248/metrics/agent
404 page not found
```

## snap with the fix

```bash
ubuntu@u1:~$ sudo cat /var/snap/maas/common/etc/maas/agent.yaml

# WARNING: MAAS will overwrite any changes made here.
...
observability:
  logging:
    level: info
  metrics:
    enabled: true
services:
...

ubuntu@u1:~$ curl http://10.226.71.227:5248/metrics/agent
# HELP cache_size
# TYPE cache_size gauge
cache_size{otel_scope_name="http_proxy",otel_scope_version="",type="current"} 0
...
```

AloizioMacedo:
When/how is this config changeable by someone/something? It seems that `agent.yaml` should not be manually changed, and we just populate it with a constant value for the metrics enabled part.

The point of the question is to understand why do we have this configuration as an option in the first place, if we want for it to always be enabled (I wasn´t able to find anything in the docs about it).

skatsaounis:
@AloizioMacedo 

Currently we are not providing a way to change it. We might want to introduce it in the future, but this is something we haven't implemented for 3.8.0. This PR though is fixing the regression. This regression is breaking the COS integration of maas-region charm, causing 404s when the opentelemetry collector is trying to scrape the endpoint. If/when we make it configurable in future MAAS releases, we will have to update this code.

AloizioMacedo (approved):

PR #429, opened 2026-07-01
Update maas-ui to c12c8b8.

dargad:


maas-lander:
- [x] Direct message
- [ ] ~maas

PR #432, opened 2026-07-02
feat: add bulk parent-child operation orchestration

tanzin8r:
Add bulk operation support in the service layer. The service creates a parent operation and starts a parent workflow; the workflow creates child rows, schedules child workflows, waits for them to finish, and rolls up the parent status to completed, failed, or completed with errors.


maas-lander:
- [x] Direct message
- [ ] ~maas

PR #434, opened 2026-07-03
fix: let normal users access their information in the v3 API

alemar99:
Resolves: LP:2148485

### DIFF
--- a/src/maasapiserver/v3/api/public/handlers/users.py
+++ b/src/maasapiserver/v3/api/public/handlers/users.py
@@ -1,7 +1,7 @@
 # Copyright 2024-2026 Canonical Ltd.  This software is licensed under the
 # GNU Affero General Public License version 3 (see the file LICENSE).
 
-from typing import Union
+from typing import Annotated, Union
 
 from fastapi import Depends, Header, Query, Response, status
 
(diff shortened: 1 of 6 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

alemar99:
QA:
- create a normal user
- in the UI go to your user page
- the page loads indefinitely
- if you open devtools, you'll see a request to `/users/{id}` which is failing

With this branch:
- in the UI go to your user page
- the page normally loads and show your user info

src/maasapiserver/v3/api/public/handlers/users.py
```diff
@@ -241,24 +241,30 @@ async def list_users(
...
+            await services.openfga_tuples.get_client().can_view_identities(
+                authenticated_user.id
+            )
+        )
         user = await services.users.get_by_id(user_id)
         if not user:
             raise NotFoundException()
 
+        if not can_view_identities and user.id != authenticated_user.id:
+            raise NotFoundException()
```
nehjoshi, line 351: Isn't 403: Forbidden more suitable here? Because then we can just hide this page in the UI and display the message, "You do not have permission to view this page."

nehjoshi (commented):
QA looks good. A question inline ->

src/maasapiserver/v3/api/public/handlers/users.py
alemar99: Nope. From a security point of view it's better to not return different status codes for different entities. A user could get, based on the status code, if the user exists (403) or if it doesn't (404)

nehjoshi (dismissed):
LGTM

src/maasapiserver/v3/api/public/models/requests/users.py
```diff
@@ -83,10 +83,6 @@ def to_builder(self) -> UserBuilder:
 
 class UserUpdateRequest(BaseUserRequest):
     password: str | None = Field(min_length=1, default=None)
-    groups: list[int] = Field(
-        default_factory=list,
-        description="The IDs of the groups the user will be a member of.",
-    )
```
nehjoshi, line 89: i.e., the "groups" form field when a viewer is viewing their own page (`/MAAS/r/account/prefs/details`) will be readonly, right?

nehjoshi (commented):
This might need to be updated too:
```src/tests/maasapiserver/v3/api/public/models/requests/test_users.py::TestUserUpdateRequest```

Also, just as an unrelated FYI, the `/users/me:change_password` isn't used anywhere in the UI. We just rely on `PUT /users/me` to update all user info

src/maasapiserver/v3/api/public/models/requests/users.py
alemar99: Yes that's the idea. If you want to edit your groups, an admin has to do it in setting > users

alemar99: And if `/users/me:change_password` is not used, should I just remove it then?

nehjoshi: It should be safe to do so, yes. The `/MAAS/a/v3/users/{user_id}:change_password` endpoint is also unused.

alemar99: Ok maybe in another commit I'd say then.

nehjoshi (approved):
LGTM!

PR #438, opened 2026-07-07
fix: use the request() method to make requests from authlib

nehjoshi:
- Authlib's `AsyncOAuth2Client` attaches tokens to requests when using `client.get()` and `client.post()`. For requests that don't require authentication (like fetching the JWKS), this is unnecessary
- Instead, we can use `client.request(withhold_token=True)` which will not add any tokens. Unfortunately, `client.get()` does not have a `withhold_token` param so we cannot call it directly.


maas-lander:
- [x] Direct message
- [ ] ~maas

alemar99 (approved):
LGTM

PR #439, opened 2026-07-08
fix: N+1 node queries in dhcp, api and websocket handlers

xtrusia:
On large subnets, several code paths run one or more node queries per object. With about 6,000 allocated IP addresses on a `/19` subnet, one DHCP configuration rebuild runs about 18,000 queries. The `ipaddresses` API has the same N+1 query problem.

On one deployment at this scale, slow API responses and deployment timeouts occurred while these DHCP configuration rebuilds were running.

This PR removes the N+1 queries from these paths and adds `count_queries()` regression tests.

- DHCP host generation in `make_hosts_for_subnets()`
- `IPAddressesHandler.read()`
- the interface listing in `Fabric.delete()`
- tracked VMs in the VM cluster handler
- `Pod.host` lookups through pod hints in the pod WebSocket handler

The fixes use `select_related()`, `prefetch_related()`, and targeted ORM queries to retrieve the required node data in a fixed number of queries.
The regression tests confirm that query counts stay constant for the covered result sets as their object counts increase.

## Performance measurements

The DHCP results below come from a source-level reproducer that calls `make_hosts_for_subnets([subnet])` directly.

1. Create a `10.42.0.0/19` subnet.
2. Create 6,000 physical interfaces and attach one AUTO `StaticIPAddress` to each interface.
3. Call `make_hosts_for_subnets([subnet])`.
4. Add the 6,001st physical interface and AUTO address with the MAAS factory.
5. Call `make_hosts_for_subnets([subnet])` again.

Query counts were collected with `connection.execute_wrapper()`. Runtimes are the median of two uninstrumented runs, with the baseline and patched results collected in the same environment.

The table shows the result after adding the 6,001st address.

| MAAS | Queries before | Queries after | Runtime before | Runtime after |
|---|---:|---:|---:|---:|
| 3.4 | 18,004 | 3 | 18.54 s | 1.25 s |
| 3.5 | 18,004 | 3 | 18.68 s | 1.32 s |
| 3.6 | 18,005 | 4 | 17.39 s | 1.18 s |
| 3.7 | 18,005 | 4 | 21.10 s | 1.81 s |
| master | 18,005 | 4 | 21.22 s | 1.47 s |

The 3.4 and 3.5 code paths do not load `ReservedIP` entries here, so the fixed count is 3. Versions 3.6, 3.7, and master also load `ReservedIP` entries, so the fixed count is 4.

The large-IP paths use `prefetch_related()`, which keeps related rows in memory and uses an `IN` clause containing the returned IP IDs. This is the memory trade-off for removing the per-IP queries.

Backports are planned for stable branches down to 3.4, where the same code paths run in production.

Resolves LP:2161952

### DIFF
--- a/src/maasserver/api/ip_addresses.py
+++ b/src/maasserver/api/ip_addresses.py
@@ -3,6 +3,7 @@
 
 """API handler: `StaticIPAddress`."""
 
+from django.db.models import Prefetch
 from django.http import Http404
 from django.http.response import HttpResponseBadRequest, HttpResponseForbidden
 from formencode.validators import StringBool
@@ -406,4 +407,12 @@ class IPAddressesHandler(OperationsHandler):
         if ip is not None:
             query = query.filter(ip=ip)
         query = query.order_by("id")
-        return query
+        # Serializing the interface_set field reads each interface's node
+        # (system_id/resource_uri call get_node()); preload the nodes so
+        # that does not run a query per IP.
+        return query.prefetch_related(
+            Prefetch(
+                "interface_set",
+                queryset=Interface.objects.select_related("node_config__node"),
+            )
+        )
--- a/src/maasserver/models/fabric.py
+++ b/src/maasserver/models/fabric.py
@@ -177,8 +177,10 @@ class Fabric(CleanSave, TimestampedModel):
                 "still present: %s" % (", ".join(descriptions))
             )
         if Interface.objects.filter(vlan__fabric=self).exists():
-            interfaces = Interface.objects.filter(vlan__fabric=self).order_by(
-                "node_config__node", "name"
+            interfaces = (
+                Interface.objects.filter(vlan__fabric=self)
+                .select_related("node_config__node")
+                .order_by("node_config__node", "name")
             )
             descriptions = [iface.get_log_string() for iface in interfaces]
             raise ValidationError(
--- a/src/maasserver/models/fabric.py
+++ b/src/maasserver/models/fabric.py
@@ -7,7 +7,15 @@ from operator import attrgetter
 import re
 
 from django.core.exceptions import PermissionDenied, ValidationError
-from django.db.models import AutoField, CharField, Manager, TextField
+from django.db.models import (
+    AutoField,
+    CharField,
+    Manager,
+    OuterRef,
+    Subquery,
+    TextField,
+)
+from django.db.models.functions import Coalesce
 from django.db.models.query import QuerySet
 
 from maasserver.fields import MODEL_NAME_VALIDATOR
@@ -177,12 +185,28 @@ class Fabric(CleanSave, TimestampedModel):
                 "still present: %s" % (", ".join(descriptions))
             )
         if Interface.objects.filter(vlan__fabric=self).exists():
-            interfaces = (
-                Interface.objects.filter(vlan__fabric=self)
-                .select_related("node_config__node")
-                .order_by("node_config__node", "name")
+            # Avoid duplicating orphan interfaces with multiple parents while
+            # preserving get_node_config()'s first-parent fallback.
+            parent_hostname = (
+                Interface.objects.filter(children=OuterRef("pk"))
+                .order_by("created")
+                .values("node_config__node__hostname")[:1]
             )
-            descriptions = [iface.get_log_string() for iface in interfaces]
+            interface_rows = (
+                Interface.objects.filter(vlan__fabric=self)
+                .annotate(
+                    resolved_hostname=Coalesce(
+                        "node_config__node__hostname",
+                        Subquery(parent_hostname),
+                    )
+                )
+                .order_by("node_config__node", "name")
+                .values_list("name", "type", "resolved_hostname")
+            )
+            descriptions = [
+                f"{name} ({interface_type}) on {hostname or '<unknown-node>'}"
+                for name, interface_type, hostname in interface_rows
+            ]
             raise ValidationError(
                 "Can't delete fabric; the following interfaces are "
                 "still connected: %s" % (", ".join(descriptions))
--- a/src/maasserver/websockets/handlers/pod.py
+++ b/src/maasserver/websockets/handlers/pod.py
@@ -6,6 +6,7 @@
 import dataclasses
 
 import attr
+from django.db.models import Prefetch
 from django.http import HttpRequest
 
 from maasserver.authorization import can_view_configurations, clear_caches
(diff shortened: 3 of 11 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

skatsaounis (changes requested):
Hi @xtrusia 

Thank you for the PR. Before starting a complete review, I wasn't able to find a linked bug report or tests that are counting queries to prove the improvements. The latter, will also stay in the code base and used as a guard for future changes - no change will be able to introduce more queries than what your changes are achieving in this code path.

Please file a bug report and add it as a footer to your PR description, then proceed with tests that are counting queries. When you are done, let us know to start the review.

This is a [link](https://github.com/canonical/maas/commit/84b734f265d99c54946b9b25b6079d0b6006947e) to a past commit that is introducing such tests. You can use it as reference.

src/maasserver/api/ip_addresses.py
```diff
@@ -406,4 +407,12 @@ def read(self, request):
...
             query = query.filter(ip=ip)
         query = query.order_by("id")
-        return query
+        # Serializing the interface_set field reads each interface's node
+        # (system_id/resource_uri call get_node()); preload the nodes so
+        # that does not run a query per IP.
+        return query.prefetch_related(
+            Prefetch(
+                "interface_set",
+                queryset=Interface.objects.select_related("node_config__node"),
```
skatsaounis, line 416: A note about this optimization. In case someone queries the IP addresses with `all=True`, in a MAAS with 10k nodes and 1 IP each:

```md
# Numbers per object where taken from a live MAAS instance
10,000 IPs * 1.97 KB = 19.7 MB
10,000 Interfaces * 3.40 KB = 34.0 MB
10,000 NodeConfigs *  1.50 KB = 15.0 MB
10,000 Nodes * 7.54 KB = 75.4 MB
Total = 144.1 MB
```

This is the footprint of just the object loading. With the serialization overhead, this could reach a 200MB spike in the memory of the region worker serving the request. I can see the benefit of pre-fetching but since it is significant I wanted to raise it here.

src/maasserver/models/fabric.py
```diff
@@ -177,8 +177,10 @@ def delete(self):
                 "still present: %s" % (", ".join(descriptions))
             )
         if Interface.objects.filter(vlan__fabric=self).exists():
-            interfaces = Interface.objects.filter(vlan__fabric=self).order_by(
-                "node_config__node", "name"
+            interfaces = (
```
skatsaounis: An alternative here is the following. This is bypassing the `Interface.get_log_string` entirely. Instead, it is reusing its string message, to avoid having to fetch the interface object in MAAS memory.

```py
if Interface.objects.filter(vlan__fabric=self).exists():
    interfaces_data = (
        Interface.objects.filter(vlan__fabric=self)
        .annotate(
            resolved_hostname=Coalesce(
                F("node_config__node__hostname"),
                F("parents__node_config__node__hostname")
            )
        )
        .order_by("resolved_hostname", "name")
        .values_list("name", "type", "resolved_hostname")
    )
    descriptions = [
        f"{name} ({iface_type}) on {hostname or '<unknown-node>'}"
        for name, iface_type, hostname in interfaces_data
    ]
    raise ValidationError(...)
```

This is what the interface methods do:

```py
    def get_node(self):
        if self.node_config is not None:
            return self.node_config.node
        return None

    def get_log_string(self):
        hostname = "<unknown-node>"
        node = self.get_node()
        if node is not None:
            hostname = node.hostname
        return f"{self.name} ({self.type}) on {hostname}"
```

And since we need only the hostname (either interface, or parent, or `<unknown-node>`), name and type, there is no need to prefetch the entire node.

src/maasserver/websockets/handlers/pod.py
```diff
@@ -75,9 +77,15 @@ class Meta:
 
     def get_queryset(self, for_list=False):
         """Return `QuerySet` for devices only viewable by `user`."""
-        return Pod.objects.get_pods(
-            self.user, PodPermission.view
-        ).select_related("hints")
+        return (
+            Pod.objects.get_pods(self.user, PodPermission.view)
+            .select_related("hints")
+            .prefetch_related(
```
skatsaounis, line 83: Note for the MAAS team:

This is the method we want to improve the performance of:

```py
def host(self):
        node = self.hints.nodes.first()
        if node:
            return node
        elif self.ip_address is not None:
            interface = self.ip_address.get_interface()
            if interface is not None:
                return interface.node_config.node
        else:
            return None
```

If a Pod lacks hints and enters the `elif self.ip_address` fallback branch, it will bypass this cache and fire lazy N+1 queries for the interface and node trees.However, given that our live Node objects are very heavy (~7.54 KB) and Interface objects are ~3.40 KB, pre-caching the entire networking tree over long-running WebSockets for an edge-case fallback path would create a permanent memory penalty. Assuming hints cover the vast majority of our active topology, leaving this as a lazy fallback is accepted. If, on the other hand, we expect the `elif` branch to be used frequently, we will have to cache its related objects too.

skatsaounis (changes requested):
Thank you for creating the bug report and introducing query count tests. This is what I was able to detect in your changes. Please take a look, especially at the 2nd comment.

r00ta (dismissed):
lgtm once @skatsaounis comments are addressed

r00ta:
It would be nice if you could grab some numbers with this new implementation so to include them in the release notes. Something like "for a /19 subnet with 6000 allocation X queries were executed; on the same environment with this fix Y queries were executed"

src/maasserver/api/ip_addresses.py
r00ta: Thanks for raising this, I think it's an acceptable compromise to do it

xtrusia: Thank you for the review!

src/maasserver/models/fabric.py
xtrusia: Thanks for the review and the suggestion. I've updated `Fabric.delete()` to build the validation message from a values-only query. For the parent hostname, I followed an existing MAAS `Subquery` pattern to preserve `get_node_config()`'s first-parent fallback for orphan interfaces and avoid duplicate rows for interfaces with multiple parents.

src/maasserver/models/fabric.py
```diff
@@ -177,10 +185,29 @@ def delete(self):
...
+                        "node_config__node__hostname",
+                        Subquery(parent_hostname),
+                    )
+                )
+                .order_by("node_config__node", "name")
+                .values_list("id", "name", "type", "resolved_hostname")
+            )
+            descriptions = [
+                f"{name} ({interface_type}) on "
+                + (hostname if hostname is not None else "<unknown-node>")
```
skatsaounis: NIT: assuming `.values_list("id", "name", "type", "resolved_hostname")` is changed to `.values_list("name", "type", "resolved_hostname")`. Is there a reason to keep "id" in the values list?

```py
descriptions = [
    f"{name} ({iface_type}) on {hostname or '<unknown-node>'}"
    for name, iface_type, hostname in interface_rows
]
```

skatsaounis (commented):
Thanks for the extra improvement. I have a nit comment, otherwise LGTM

r00ta (approved):

skatsaounis (approved):
Nice work! LGTM

PR #443, opened 2026-07-08
fix: make rsyslog formatting for tcp wrapping match rfc6587 specification

MaikRe:
Resolves LP:2159044


maas-lander:
- [ ] Direct message
- [ ] ~maas

r00ta (approved):

PR #445, opened 2026-07-09
fix: let normal users access their information in the v3 API

alemar99:
Resolves: LP:2148485


maas-lander:
- [ ] Direct message
- [ ] ~maas

alexsander-souza (approved):
+1

PR #448, opened 2026-07-09
docs: link feedback button to launchbad bug report rather than nonexistent github issues page

MaikRe:


maas-lander:
- [ ] Direct message
- [ ] ~maas

AloizioMacedo:
We should ideally not change this `overwrite_links.js`, at least too much. It is intended to replace at runtime the links to not use the read the docs ones, and instead use the `canonical.com/maas/docs` ones.

This seems to me to be a feature that should be done in the `canonical-sphinx` package itself. Looking into it, it seems to hardcode the assumption that github is used to track issues: https://github.com/canonical/canonical-sphinx/blob/f7c6a4da8897e705b742670459cf7c7921bf0af3/canonical_sphinx/theme/static/github_issue_links.js#L17

Can you maybe reach out to them and ask for allowing users to set up a direct link to the issues, instead of trying to concoct one from the github url as above? (And confirm that this is not possible currently?)

MaikRe:
> We should ideally not change this `overwrite_links.js`, at least too much. It is intended to replace at runtime the links to not use the read the docs ones, and instead use the `canonical.com/maas/docs` ones.

I have update this to completely overwrite the github issue links javascript file instead as was recommended to me by the docs team until this is full supported in the template.

It is a known issue: https://github.com/canonical/canonical-sphinx/issues/51 but as I have it now will be sufficient until a fix is implemented.

AloizioMacedo:
> > We should ideally not change this `overwrite_links.js`, at least too much. It is intended to replace at runtime the links to not use the read the docs ones, and instead use the `canonical.com/maas/docs` ones.
> 
> I have update this to completely overwrite the github issue links javascript file instead as was recommended to me by the docs team until this is full supported in the template.
> 
> It is a known issue: [canonical/canonical-sphinx#51](https://github.com/canonical/canonical-sphinx/issues/51) but as I have it now will be sufficient until a fix is implemented.

Nice! Thanks, @MaikRe !

Before merging, let's add some reference to https://github.com/canonical/canonical-sphinx/issues/51 as a comment above the `html_js_files` list, explaining why that file is there, to reduce the likelihood that we just forget to update when they push the change.

AloizioMacedo (approved):

PR #450, opened 2026-07-10
3.4.10 re release #2

goldberl:
Moving release commit again to include cherry-pick for replace deprecated pkg_resources fix.


maas-lander:
- [x] Direct message
- [ ] ~maas

PR #455, opened 2026-07-13
chore: upgrade Go to 1.24

alexsander-souza:


maas-lander:
- [x] Direct message
- [ ] ~maas

PR #457, opened 2026-07-13
docs: document switch installer syslog logging

tanzin8r:


maas-lander:
- [x] Direct message
- [ ] ~maas

docs/how-to-guides/manage-switches.md
```diff
@@ -211,6 +211,100 @@ MAAS automatically handles the DHCP configuration needed for ONIE. When a switch
...
+   syslogd -b 3 -D -L -R "${RACK_IP}:${SYSLOG_PORT}"
+
+   logger -t nos-installer -p local0.info "wrapper start"
+   # Run your NOS installer here
+   logger -t nos-installer -p local0.info "wrapper finished"
+   ```
+
+   The `-b 3 -D -L` flags match ONIE defaults. `-R` adds the remote MAAS target.
+
+4. **Check the logs on a region controller** (or a combined region+rack host):
```
MaikRe: Can you link to the general how-to-logging section here? I believe the commands for snap/deb differ slightly for checking logs

docs/how-to-guides/manage-switches.md
```diff
@@ -211,6 +211,100 @@ MAAS automatically handles the DHCP configuration needed for ONIE. When a switch
...
+
+   ```bash
+   journalctl -f SYSLOG_IDENTIFIER=maas-machine | grep --line-buffered nos-installer
+   ```
+
+💡 Always include the port in `-R <host>:<port>`. If you omit it, BusyBox defaults to UDP port `514`, which is not `maas_syslog_port`.
+
+💡 BusyBox `logger` only supports `-s`, `-t`, and `-p`. It writes to the local `syslogd` and cannot open a remote host by itself.
+
+💡 On a rack-only controller, MAAS receives remote messages and forwards them to the region; it does not write them to the local journal. After forwarding, `MAAS_MACHINE_IP` is the rack, so prefer the tag filter.
```
MaikRe: While I personally don't mind the use of emojis here, I am not sure if this matches the style of the rest of the documentation. If it does please let me know and keep it as is, if not it might be better to do something like:

#### Important Notes
- Always include ..
- BusyBox 'logger' ...
- On a rack-only ...

docs/how-to-guides/manage-switches.md
```diff
@@ -211,6 +211,100 @@ MAAS automatically handles the DHCP configuration needed for ONIE. When a switch
...
+   ```bash
+   maas $PROFILE subnets read
+   maas $PROFILE dhcpsnippets create \
+     name=onie-log-servers \
+     subnet=<subnet-id> \
+     enabled=1 \
+     value='option log-servers <syslog-server-ip>;'
+   ```
+
+   Equivalent ISC syntax:
```
MaikRe: Is this meant to be for the DHCP snippet? If so please include that in the description here as well

docs/how-to-guides/manage-switches.md
```diff
@@ -211,6 +211,100 @@ MAAS automatically handles the DHCP configuration needed for ONIE. When a switch
...
+
+   ```text
+   option log-servers <syslog-server-ip>;
+   ```
+
+3. **Reboot the switch into ONIE** (or renew DHCP in ONIE) so it picks up option 7.
+
+4. **Confirm messages on your collector.** They will not appear in the MAAS journal.
+
+💡 DHCP snippets are deprecated as of MAAS 3.6 and will be removed in the next major version. Prefer an external DHCP server for this option when you can. See [Manage network services](manage-network-services.md) and [ONIE: Add a Remote Syslog Server](https://opencomputeproject.github.io/onie/user-guide/index.html#add-a-remote-syslog-server).
```
MaikRe: Same note regarding emoji usage

MaikRe (changes requested):

MaikRe (approved):
+1

PR #458, opened 2026-07-13
fix: use name of interface instead of relying on idx

AloizioMacedo:
When creating a machine, we do some song and dance with LXD and the interfaces that are specified seem to be retrieved in alphabetical order of their names, even though we assume same-order of the listing. This is likely an artifact of the fact that LXD is written in Go, and [marshalling sorts keys](https://pkg.go.dev/encoding/json).

We now store and use the name directly later, if available. It may not be available since that same code path is run for virsh as well.

Resolves: [LP:2160526](https://bugs.launchpad.net/maas/+bug/2160526)

### DIFF
--- a/src/maasserver/models/bmc.py
+++ b/src/maasserver/models/bmc.py
@@ -1107,8 +1107,16 @@ class Pod(BMC):
         # configuration.
         created_interfaces = []
         for idx, discovered_nic in enumerate(discovered_machine.interfaces):
+            # For the scenario of lxd, it returns the interfaces ordered
+            # alphabetically likely due to it being written in Go. So we need
+            # to properly match the correct name here instead of relying on
+            # order.
+            # When coming from virsh, the interfaces do not have a meaningful
+            # name. So we fall back to the previous behavior (which might
+            # be even correct in that scenario.)
+            name = discovered_nic.name or interface_names[idx]
             interface = self._create_interface(
-                discovered_nic, machine, name=interface_names[idx]
+                discovered_nic, machine, name=name
             )
             created_interfaces.append(interface)
             if discovered_nic.boot:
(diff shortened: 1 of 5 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

src/provisioningserver/drivers/pod/tests/test_lxd.py
```diff
@@ -995,6 +995,7 @@ def test_get_discovered_machine(self):
                 boot=True,
                 attach_type=InterfaceAttachType.BRIDGE,
                 attach_name="lxdbr0",
+                name="eth0",
```
tanzin8r, line 998: consider adding a regression test as well. since these assertions confirm discovery populates `name` when devices are already eth0..ethN (insertion order = alpha order).

src/maasserver/models/bmc.py
tanzin8r, line 1090: the bit about matching hypervisor order is a bit misleading now for LXD.

src/provisioningserver/drivers/pod/lxd.py
tanzin8r, line 763: just curious about this. we set boot.priority on the first requested NIC at compose time, but discovery marks boot from list order here. With custom names, would that still be the same NIC?

tanzin8r (changes requested):
left a couple of comments.

src/maasserver/models/bmc.py
AloizioMacedo: It was misleading before too : P, since LXD did not actually respect the order.

But good point! Will push a change.

AloizioMacedo:
This fix does not apply to master anymore, given the changes in LXD VM Host support. I will update it to target 3.8 later, but it will have to wait due to the 3.8 release process.

PR #453, opened 2026-07-13
fix: properly calculate machine resources when bcache is configured

skatsaounis:
Bump go.mod of host-info binary. Doing so, will allow MAAS properly conduct hardware sync on machines with bcache. Without the bump, the LXD code that host-info consumes fails to successfully return when block devices are formulating a bcache. Relevant LXD fix: https://github.com/canonical/lxd/commit/57c38b760253cfc68a71c905426621599090ca0d

Resolves: LP#2160547

PR #461, opened 2026-07-14
fix: properly calculate machine resources when bcache is configured (#453)

skatsaounis:
Bump go.mod of host-info binary. Doing so, will allow MAAS properly conduct hardware sync on machines with bcache. Without the bump, the LXD code that host-info consumes fails to successfully return when block devices are formulating a bcache. Relevant LXD fix: https://github.com/canonical/lxd/commit/57c38b760253cfc68a71c905426621599090ca0d

Resolves: LP#2160547
(cherry picked from commit 6f99eed37b10c78dda491b548cffba328a7aff83)

### DIFF
--- a/src/host-info/go.mod
+++ b/src/host-info/go.mod
@@ -2,7 +2,7 @@ module host-info
 
 go 1.26.4
 
-require github.com/canonical/lxd v0.0.0-20260619131343-14db131f6598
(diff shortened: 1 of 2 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

PR #463, opened 2026-07-14
feat(security): FIPS 140-2/140-3 and STIG/CIS compliance

alexsander-souza:


maas-lander:
- [x] Direct message
- [ ] ~maas

snap/snapcraft.yaml
```diff
@@ -20,8 +20,7 @@ epoch: 4*
 
 package-repositories:
   - type: apt
-    ppa: maas/3.7-next
-
+    ppa: alexsander-souza/maas-3.7-fips
```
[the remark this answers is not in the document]
alexsander-souza: temporary, should be reverted once the packages land in `3.7-next`

snap/local/tree/bin/run-named
```diff
@@ -16,13 +16,6 @@ MAAS_ZONE_FILE_CONFIG_DIR="$SNAP_DATA/bind" \
     MAAS_DNS_CONFIG_DIR="$SNAP_DATA/bind" \
     "$SNAP/bin/maas-rack" setup-dns --no-clobber
 
-# named default socket limit (21k) can be higher than current limits.
-# Increase the limit to the max allowed and restrict named to this.
-# LP:1901999
-ulimit -n "$(ulimit -Hn)"
-MAAS_DNS_SOCK_LIMIT=$(ulimit -n)
-
 exec "$SNAP/usr/sbin/named" \
```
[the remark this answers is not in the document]
alexsander-souza: this parameter is a no-op in recent versions

src/maascommon/hardening.py
```diff
@@ -0,0 +1,52 @@
...
+from maascommon.fips import is_fips_enabled
+
+_logger = logging.getLogger("maas.hardening")
+
+#: Set by configure_hardening() at process startup.
+_hardening_active: bool = False
+_hardening_configured: bool = False
+
+
+def configure_hardening(hardening_enabled: "str | None") -> None:
```
AloizioMacedo:
```suggestion
def configure_hardening(hardening_enabled: Literal["auto", "on", "off"] | None) -> None:
```

src/maascommon/password_policy.py
```diff
@@ -0,0 +1,65 @@
...
+"""Password and credential complexity validators for MAAS hardening."""
+
+from dataclasses import dataclass, field
+import re
+
+
+@dataclass
+class PasswordValidationResult:
+    is_valid: bool
+    errors: list[str] = field(default_factory=list)
```
AloizioMacedo, line 11: No need to allow a invalid state (`PasswordValidationResult(is_valid=False, errors=[])` for example)

You can make `is_valid` a method of `PasswordValidationResult` that just `return not self.errors`.

src/maasserver/djangosettings/settings.py
```diff
@@ -98,6 +98,22 @@ def _get_default_db_config(config: RegionConfiguration) -> dict:
...
+            **(
+                {"sslkey": config.database_sslkey}
+                if config.database_sslkey
+                else {}
+            ),
+            **(
+                {"sslrootcert": config.database_sslrootcert}
+                if config.database_sslrootcert
+                else {}
+            ),
```
AloizioMacedo: Maybe personal, but this feels a bit too clever. I think with the same number of lines (maybe less?) we can just create 

```python
        options= {
            "keepalives": int(config.database_keepalive),
            "keepalives_idle": config.database_keepalive_idle,
            "keepalives_interval": config.database_keepalive_interval,
            "keepalives_count": config.database_keepalive_count,
            "application_name": application_name,
            "sslmode": config.database_sslmode,
        }
```

and do the checks for each one of the three optional configs, adding the keys to options when relevant, e.g. 
```python
    if config.database_sslkey:
        options["sslkey"] = config.database_sslkey
``` 

and returning `{..., "OPTIONS":  options}`

src/maasserver/models/config.py
```diff
@@ -205,3 +206,41 @@ def ensure_uuid_in_config() -> str:
...
+            )
+    except Exception:
+        logging.getLogger(__name__).warning(
+            "Could not read hardening_enabled from DB; defaulting to None.",
+        )
+        return None
+    return str(value) if value is not None else None
+
+
+def read_fips_declared_from_db() -> "bool | None":
```
AloizioMacedo: We don't need the stringified type. Some other functions are also using them, but we can just remove the quotes. Stringfying the type in our Python version is mainly used for some out-of-order type referencing.

```suggestion
def read_fips_declared_from_db() -> bool | None:
```

src/maasserver/start_up.py
```diff
@@ -251,6 +264,10 @@ def inner_start_up(master=False):
     # this early, only on the master process, to ensure they exist and are
     # only created once. If get_or_create_running_controller() is called before
     # this it will fail on first run.
+    configure_hardening(read_hardening_enabled_from_db())
```
AloizioMacedo, line 262: I think the comment just above referred to the `if master:` block, so we might want to move it down.

src/maasservicelayer/db/repositories/power_types.py
```diff
@@ -0,0 +1,43 @@
...
+        for pt in power_types:
+            entry = dict(pt)
+            name = entry.get("name", "")
+            status, reason = DRIVER_FIPS_REGISTRY.get(
+                name, (DriverFIPSStatus.COMPLIANT, None)
+            )
+            entry["fips_supported"] = status == DriverFIPSStatus.COMPLIANT
+            entry["fips_unsupported_reason"] = reason
+            result.append(entry)
+        return result
```
AloizioMacedo, line 43: Why is list async here? I saw that there are some async `list` methods that do some pagination queries in `ReadOnlyRepository" for example , but the abstraction does not seem to be shared between the types, and async seems unnecessary here.

src/maasapiserver/v3/api/public/handlers/system.py
```diff
@@ -0,0 +1,48 @@
...
+from maasapiserver.v3.api import services
+from maasapiserver.v3.auth.base import check_permissions
+from maascommon.fips import get_fips_status
+from maasservicelayer.auth.jwt import UserRole
+from maasservicelayer.services import ServiceCollectionV3
+
+
+class SystemInfoResponse(BaseModel):
+    fips_active: bool
+    version: str
```
MaikRe, line 19: The spec lists both fips_active and hardening active as required outputs of the system info response.
Why is this not included here?

src/maasapiserver/v3/api/public/handlers/system.py
```diff
@@ -0,0 +1,48 @@
...
+
+    TAGS = ["System"]
+
+    @handler(
+        path="/system/info",
+        methods=["GET"],
+        tags=TAGS,
+        responses={200: {"model": SystemInfoResponse}},
+        dependencies=[
+            Depends(check_permissions(required_roles={UserRole.USER}))
```
MaikRe, line 33: The spec states "The /system/info endpoint requires no authentication as it exposes only non-sensitive operational state." Either the spec needs to include the fact that a logged in user is required or this permissions check must be removed.

src/maasservicelayer/services/hardening.py
```diff
@@ -0,0 +1,349 @@
...
+
+        try:
+            dh_params = load_pem_parameters(dhparam_path.read_bytes())
+            bit_length = dh_params.parameter_numbers().p.bit_length()
+            if bit_length < 2048:
+                return [
+                    _violation(
+                        code="WEAK_DH_PARAMS",
+                        message=f"DH parameters are {bit_length} bits; minimum is 2048",
+                        resolution="Run: openssl dhparam -out dhparam.pem 2048, then set api_tls_dhparam in regiond.conf",
```
MaikRe: It seems weird to me that editing the regiond.conf manually is a recommended resolution step. I thought the goal was to avoid having operators ever edit this file manually?

src/maascommon/logging/security.py
```diff
@@ -0,0 +1,92 @@
...
+    _log.info(
+        "%s",
+        FIPS_TLS_HANDSHAKE,
+        extra={
+            "event": FIPS_TLS_HANDSHAKE,
+            "cipher_suite": cipher_suite,
+            "protocol_version": protocol_version,
+            "peer": peer,
+            "cert_issuer": cert_issuer,
+            "verify_enabled": verify_enabled,
```
MaikRe: There is a discrepancy with the spec here. 
The spec lists `cert_valid` as one of the logging items and does not list `verify_enabled`.

src/maasapiserver/common/api/models/responses/errors.py
```diff
@@ -103,14 +106,25 @@ def __init__(self, details: Optional[list[BaseExceptionDetail]]):
...
+        self,
+        details: Optional[list[BaseExceptionDetail]],
+        fips_violation: Optional[bool] = None,
+    ):
+        body_response = ValidationErrorBodyResponse(
+            details=details,
+            fips_violation=fips_violation,
+        )
+        encoded = jsonable_encoder(body_response)
+        # Strip null fips_violation so normal 422s stay clean.
```
MaikRe: redundant comment

src/maasserver/forms/fips_power.py
```diff
@@ -0,0 +1,82 @@
...
+    if not supported:
+        alternatives = get_fips_compliant_alternatives()
+        raise ValidationError(
+            f"Power driver '{power_type}' is not supported in FIPS mode: "
+            f"{reason}. FIPS-compliant alternatives: "
+            f"{', '.join(sorted(alternatives))}",
+            code="fips_violation",
+        )
+
+    # IPMI: enforce cipher suite 17
```
MaikRe: redundant comment

src/maasserver/forms/fips_power.py
```diff
@@ -0,0 +1,82 @@
...
+) -> None:
+    """Raise ValidationError if power parameters violate FIPS requirements.
+
+    Called from AdminMachineForm.clean() and PodForm.clean() when FIPS is
+    enabled on the host.  Does nothing when FIPS is not enabled.
+    """
+    if not is_fips_enabled():
+        return
+
+    # Check driver FIPS support
```
MaikRe: redundant comment

MaikRe (changes requested):
A few questions inline, plus I marked some redundant comments. There are more in the code but I have only marked some of them as removing them would only be for code cleanliness and has no effect on the implemented logic.

src/maasapiserver/v3/api/public/handlers/system.py
alexsander-souza: Fixed — `hardening_active: bool` has been added to `SystemInfoResponse` and is populated via `is_hardening_enabled()`. The response now includes both `fips_active` and `hardening_active` as the spec requires.

src/maasapiserver/v3/api/public/handlers/system.py
alexsander-souza: The endpoint requires authentication — the spec needs to be updated to reflect this

MaikRe (approved):

PR #464, opened 2026-07-14
docs: add credential injection instructions by means of sonic community example

MaikRe:


maas-lander:
- [ ] Direct message
- [ ] ~maas

alexsander-souza (approved):

PR #467, opened 2026-07-15
fix(network): fix race conditions in neighbour table updates 

alexsander-souza:
Two concurrency bugs caused spurious DB errors under high ARP traffic:

update_neighbour() loaded a Neighbour row and then saved it in the same
transaction, triggering REPEATABLE READ conflicts when a concurrent
transaction deleted or updated the same row. Replace with a single
round-trip UPDATE using F("count")+1. If the row is deleted between
get_or_create and the UPDATE the observation is silently dropped, which
is acceptable for best-effort ARP data.

delete_and_log_obsolete_neighbours() queried previous bindings without
taking a row-level lock, allowing a concurrent writer to modify those
rows between the SELECT and the DELETE. Add select_for_update() to
serialise that window.

Once the @transactional retry budget is exhausted, retryable
DatabaseErrors (serialisation failures, deadlocks) propagated as
critical log entries from the ReportNeighbours RPC handler. Since
neighbour observations are best-effort, suppress those errors after
retries are exhausted and log at warning level instead. Non-retryable
DatabaseErrors still propagate so genuine bugs remain visible.

### DIFF
--- a/src/maasserver/models/interface.py
+++ b/src/maasserver/models/interface.py
@@ -1686,6 +1686,10 @@ class Interface(CleanSave, TimestampedModel):
             # Use a single-round-trip UPDATE to avoid a REPEATABLE READ
             # conflict: loading the row and then saving it in the same
             # transaction races with concurrent deletions or updates.
+            # If the row was deleted by a concurrent transaction after
+            # get_or_create returned it, this UPDATE matches 0 rows and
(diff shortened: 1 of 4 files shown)


maas-lander:
- [ ] Direct message
- [ ] ~maas

wyattrees (approved):
lgtm

PR #468, opened 2026-07-15
feat: add hardware profile table, repository and service

alemar99:


maas-lander:
- [ ] Direct message
- [ ] ~maas

motrilense (approved):
lgtm

Some CI tests failed because the feature branch makes the job to use an incorrect docker image. Maybe it is better fixing it before adding more PRs

alemar99:
I've built the docker image and now CI seems running, thanks for the heads up!

PR #469, opened 2026-07-15
chore: bump snapcraft epoch

r00ta:
Bump the snapcraft epoch to 5 for MAAS 3.8 LTS. This way, we can start cleaning up some of logics we have around migrations of existing environments (django migrations, moving images from DB to the disk and others)


maas-lander:
- [ ] Direct message
- [ ] ~maas

bjornt (approved):
+1

PR #471, opened 2026-07-15
chore: bump snapcraft to epoch 6

r00ta:
Bump snapcraft epoch to version 6, where we drop image migration from the database and django migrations.


maas-lander:
- [ ] Direct message
- [ ] ~maas

alemar99 (approved):

PR #473, opened 2026-07-15
refactor: remove django migrations and db->disk image migrations

r00ta:
Since MAAS 3.8 (epoch 5) is a mandatory upgrade hop before moving to MAAS 4.0 (epoch 6, snap-only), we can safely remove outdated migration paths.

Changes include:
- Removed all legacy Django migrations
- Removed image migration logic


maas-lander:
- [ ] Direct message
- [ ] ~maas

alexsander-souza (dismissed):
Image

alexsander-souza (approved):

PR #474, opened 2026-07-15
fix(dns): remove deprecated named -S socket-limit flag (#466)

alexsander-souza:
The -S flag and the reserved-sockets mechanism were deprecated in BIND 9 and cause warnings or errors on newer releases. The ulimit workaround for LP:1901999 is no longer needed as BIND now handles socket limits internally.


maas-lander:
- [ ] Direct message
- [ ] ~maas

PR #492, opened 2026-07-17
feat: Add name column to switch table 3.8

MaikRe:


MaikRe:
Existing environments will need the following change applied to their database:
```sql
BEGIN;
ALTER TABLE maasserver_switch
ADD COLUMN IF NOT EXISTS name VARCHAR(255) NULL;
COMMIT;
```

PR #494, opened 2026-07-20
feat: Return mac address in switch response

MaikRe:


nehjoshi (approved):
+1, thank you for this change!

PR #496, opened 2026-07-20
fix: fix boot resources race condition

AloizioMacedo:
We are having several issues in our testplan runs in SolQA due to race conditions regarding file deletions.

This PR checks, just before deleting, if the files are being referenced. If they are, keep them.

It would perhaps be more ideal to implement some sort of "garbage collection" instead of just avoiding deletion and relying on proper triggers later to delete the files in case they become stale, but this would be a much bigger change and it is unclear if it would bring much more value.

Worth noticing: this race condition was present before 3.8. It is also present in 3.5 (see here for more details: https://bugs.launchpad.net/maas/+bug/2152843), but apparently we are hitting it much more frequently now, perhaps due to infrastructure issues.

### DIFF
--- a/src/maastemporalworker/workflow/bootresource.py
+++ b/src/maastemporalworker/workflow/bootresource.py
@@ -940,7 +940,7 @@ class DeleteBootResourceWorkflow:
     The files targeted by this workflow are chosen when it is scheduled,
     but execution (particularly resolving region endpoints) can be
     delayed. If a newer import recreates a BootResourceFile with the same
-    sha256 the meantime, deleting it would remove content that is genuinely
+    sha256 in the meantime, deleting it would remove content that is genuinely
     in use again. So this workflow re-checks the database immediately
     before deleting and skips any file that is no longer safe to remove.
     """
--- a/src/maastemporalworker/workflow/bootresource.py
+++ b/src/maastemporalworker/workflow/bootresource.py
@@ -348,8 +348,34 @@ class BootResourcesActivity(ActivityBase):
     async def delete_bootresourcefile(
         self, param: ResourceDeleteParam
     ) -> bool:
-        """Delete files from disk"""
+        """Delete files from disk, unless they have become referenced again.
+
+        The files to delete are chosen when the workflow is scheduled, but
+        execution (particularly resolving region endpoints) can be
+        delayed. If a newer import recreates a BootResourceFile with the
+        same sha256 in the meantime, deleting it would remove content that
+        is genuinely in use again. So each file is re-checked against the
+        database immediately before it is unlinked.
+        """
+        async with self.start_transaction() as services:
+            still_referenced_shas = {
+                f.sha256
+                for f in await services.boot_resource_files.get_many(
+                    query=QuerySpec(
+                        where=BootResourceFileClauseFactory.with_sha256_in(
+                            [file.sha256 for file in param.files]
+                        )
+                    )
+                )
+            }
         for file in param.files:
+            if file.sha256 in still_referenced_shas:
+                logger.info(
+                    f"skipping delete of {file}: a BootResourceFile with "
+                    "this sha256 was recreated since the deletion was "
+                    "scheduled"
+                )
+                continue
             logger.debug(f"attempt to delete {file}")
             lfile = AsyncLocalBootResourceFile(
                 file.sha256, file.filename_on_disk, 0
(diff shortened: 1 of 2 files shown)


src/maastemporalworker/workflow/bootresource.py
```diff
@@ -348,8 +348,37 @@ async def download_bootresourcefile(
...
+        delayed. If a newer import recreates a BootResourceFile with the
+        same sha256 in the meantime, deleting it would remove content that
+        is genuinely in use again. So each file is re-checked against the
+        database immediately before it is unlinked.
+        """
+        if not param.files:
+            return True
+
+        async with self.start_transaction() as services:
+            still_referenced_shas = {
```
MaikRe, line 364: You may still have a time-of-reference/time-of-use issue here.
What happens if a file is not reference and thus still_reference_shas is empty but between this call and when the file is actually unlinked a new reference to it is created?

src/maastemporalworker/workflow/bootresource.py
```diff
@@ -906,7 +935,15 @@ async def run(self, input: SyncRequestParam) -> None:
...
 @workflow.defn(name=DELETE_BOOTRESOURCE_WORKFLOW_NAME, sandboxed=False)
 class DeleteBootResourceWorkflow:
-    """Delete a BootResourceFile from this cluster"""
+    """Delete a BootResourceFile, unless it has become referenced again.
+
+    The files targeted by this workflow are chosen when it is scheduled,
+    but execution (particularly resolving region endpoints) can be
+    delayed. If a newer import recreates a BootResourceFile with the same
+    sha256 in the meantime, deleting it would remove content that is genuinely
+    in use again. So this workflow re-checks the database immediately
```
MaikRe: Minor nit but I believe the activity technically rechecks the database not the workflow. However I don't think this really changes the overall meaning of whats being done

src/maastemporalworker/workflow/bootresource.py
```diff
@@ -916,15 +953,18 @@ async def run(self, input: ResourceDeleteParam) -> None:
...
                 DELETE_BOOTRESOURCEFILE_ACTIVITY_NAME,
                 input,
                 task_queue=f"region:{r}",
                 start_to_close_timeout=DISK_TIMEOUT,
                 schedule_to_close_timeout=DISK_TIMEOUT,
                 retry_policy=RetryPolicy(maximum_attempts=3),
             )
+            for r in regions
+        ]
+        await asyncio.gather(*tasks)
```
MaikRe, line 968: This changes the behavior of the workflow from sequential to parallel. Do any tests need to be updated to validate that the workflows are correctly scheduled and more importantly what the expected behavior is when one or more of them fail?

MaikRe (commented):
Some comments/nits/questions inline

alexsander-souza (dismissed):

src/maastemporalworker/workflow/bootresource.py
AloizioMacedo: This is a good point and I thought about not including this change together in this PR.

However, it directly helps with mitigating the race condition, since the race condition is really mostly a timeout issue, so by reducing the time until deletion we are helping to avoid it directly. (As a sidenote, I am suspecting that we are seeing this appear in the test because of infrastructure issues which might be related to PS7 and be causing uncommon delays.)

The workflows being correctly scheduled is not an issue (concurrency of the scheduling itself does not change anything about that). Regarding the behavior itself, to simplify it a lot, this delete activity is essentially: 

1. Check the DB for references of a given file
2. Delete the file **from the region**

Note that it doesn't write to the db, it only observes it. And the mutability happens on the regions, which are isolated. (The problem that some region might be looking at a different version of the DB than another is actually **mitigated** with this change, due to what I explained before.)

src/maastemporalworker/workflow/bootresource.py
AloizioMacedo: That is true, but getting rid of a possible race condition entirely would be more complicated and I don't think it is worth it (for the purpose of having a fix for the release), as the race condition you refer to is orders of magnitude less likely.

The point is that the race condition issue that the PR is trying to solve is really not a precision problem. The errors that were seen in the test runs are due to the fact that the time between the scheduling and the actual deletion is taking around 2 min, which I understand to be already unusual.

Copilot tells me it is taking that time likely due to a static IP not being resolvable yet. But regardless, the window for the timeout problem is going from 2 min (time from scheduling to actual deletion) to a matter of milliseconds (time to unlink the file + some other minor things).

src/maastemporalworker/workflow/bootresource.py
AloizioMacedo: The activity is part of the workflow, so it also technically rechecks the db.

I will rephrase it in a way that accommodates both of our points.

src/maastemporalworker/workflow/bootresource.py
AloizioMacedo: Oh, and I forgot to respond regarding the "failure" part.

Well, the behavior before is arguably weirder, since a failure there would favor regions in order. There is no error handling anyway, so the material behavioral change is that here, they would fail independently.

src/maastemporalworker/workflow/bootresource.py
MaikRe: Fair enough, I did not consider the time scales involved here :+1:

MaikRe (approved):
+1

PR #504, opened 2026-07-23
chore: MAAS as a snap only - drop debian packaging

r00ta:
MAAS 4.0 will be released only as a snap. This commit removes all the debian packaging code and conditional branches in the services.


troyanov (approved):
LGTM ❤️

src/provisioningserver/utils/arch.py
```diff
@@ -28,7 +28,7 @@ def get_architecture():
     """Get the Debian architecture of the running system."""
     arch = os.getenv("SNAP_ARCH")
     if not arch:
-        # assume it's a deb environment
+        # not running in a snap
```
alexsander-souza: should not happen

src/provisioningserver/utils/__init__.py
```diff
@@ -107,11 +105,12 @@ def _flatten(things):
 
 
 def sudo(command_args):
```
alexsander-souza: this method is no-op, we could remove it from everywhere

src/provisioningserver/utils/fs.py
```diff
@@ -289,67 +260,19 @@ def incremental_write(content, filename, mode=0o600, uid=None, gid=None):
...
-
-    if is_dev_environment():
-        from maastesting import dev_root
-
-        interpreter = os.path.join(dev_root, "bin", "py")
-        command = interpreter, *command
-    return command
-
-
 def sudo_write_file(filename, contents, mode=0o644):
```
alexsander-souza: short-circuit all calls to atomic_write()

src/provisioningserver/utils/fs.py
```diff
@@ -289,67 +260,19 @@ def incremental_write(content, filename, mode=0o600, uid=None, gid=None):
...
-        if not is_dev_environment():
-            command = sudo(command)
-        proc = Popen(command, stdin=PIPE)
-        stdout, stderr = proc.communicate(contents)
-        if proc.returncode != 0:
-            raise ExternalProcessError(proc.returncode, command, stderr)
+    atomic_write(contents, filename, mode=mode)
 
 
 def sudo_delete_file(filename):
```
alexsander-souza: short-circuit calls to atomic_delete()

src/provisioningserver/utils/version.py
```diff
@@ -85,7 +83,7 @@ def extended_info(self) -> str:
     @classmethod
     def from_string(cls, version: str):
         r = re.compile(
-            r"((?P<epoch>\d+):)?"  # deb package epoch
+            r"((?P<epoch>\d+):)?"  # optional version epoch
```
alexsander-souza: harmless, but it will never match

Makefile
```diff
@@ -96,7 +96,6 @@ install-dependencies:
 ifneq ($(MAAS_PPA),)
 	sudo apt-add-repository -y $(MAAS_PPA)
 endif
-	$(apt) build-dep .
```
alexsander-souza, line 99: `golang-go`, `python3-django` and `python3-twisted` were installed here, we need to move this to `required-packages/base` or `required-packages/dev`

debian/changelog
alexsander-souza, line 1: IIRC this was the central point that relates a MAAS release to its target Ubuntu release, and there are a number of workflows and Jenkins jobs that read this file to figure out which Ubuntu release to use.

what's the plan now?

src/maastesting/pytest/environment.py
```diff
@@ -30,6 +30,22 @@ def setup_testenv(monkeypatch, tmpdir):
...
+    # required by snap-only code paths (e.g. get_running_version,
+    # get_maas_cert_tuple) that would otherwise crash when unset. ``SNAP``
+    # (the snap root path) is deliberately left unset: setting it would change
+    # path-prefixing behaviour (curtin helpers, nginx static roots, wsman
+    # config) and break tests exercising the default paths.
+    snap_common = tmpdir.join("snap_common")
+    snap_common.mkdir()
+    monkeypatch.setenv("SNAP_COMMON", str(snap_common))
+    monkeypatch.setenv("SNAP_VERSION", "3.0.0-456-g.deadbeef")
+    monkeypatch.setenv("SNAP_REVISION", "1234")
```
alexsander-souza, line 43: set `SNAP_ARCH` also

src/provisioningserver/dns/config.py
```diff
@@ -289,7 +278,9 @@ def set_up_rndc():
     """
     rndc_content, named_content = generate_rndc(
         port=get_dns_rndc_port(),
-        include_default_controls=get_dns_default_controls(),
+        # The default controls don't work in a confined snap, since it
+        # implicitly requires access to /etc/bind
+        include_default_controls=False,
```
alexsander-souza: drop this parameter (and related code branch) from `generate_rndc()`

Makefile
r00ta: I've added `golang-go` to `required-packages/dev`, but `python3-django` and `python3-twisted` seem to be already installed with `make install-dependencies` without any addition

r00ta: they're transitive dependencies for other packages such as `python3-django-piston3`. I'll make them explicit anyways

debian/changelog
r00ta: I believe we'll just move away from it as we "only" have to care about snap channels from now on. We don't need anymore specific commits to be added when we prepare a release, we only need a tag on the commit that passed QA and that we're going to release. wdyt?

alexsander-souza (dismissed):
+1

alexsander-souza (dismissed):

nehjoshi (approved):
lgtm 🚀

PR #505, opened 2026-07-23
feat: add config column to OIDCProviderTable

nehjoshi:
- This PR adds a `config` column to the `OIDCProviderTable`, which will be used to provide any additional config parameters for setting up the provider. For e.g., since we have requests to support Google as an IdP, we require additional attributes (such as `client_email`, `private_key` etc.) for the provider. These attributes can differ between providers, thus the column is of the type `JSONB`. Some providers that are already supported (e.g. Entra ID) do not require additional config params, and thus, this column is nullable.
- The goal is that these additional config options will be shown to the user in the UI/API, and from there we can parse and store this information in an appropriate way in the table.
- This will also be backported to the `3.8` branch so that the migration is in place, and work for integrating Google will begin later.

### DIFF
--- a/src/maasservicelayer/builders/external_auth.py
+++ b/src/maasservicelayer/builders/external_auth.py
@@ -2,6 +2,7 @@
 # GNU Affero General Public License version 3 (see the file LICENSE).
 
 from datetime import datetime
+from typing import Any
 
 from pydantic import Field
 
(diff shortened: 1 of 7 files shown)


r00ta (commented):
are you planning to backport to 3.8?

r00ta (changes requested):
Sorry, I just read the description. If you plan to backport to 3.8 you can't use this migration as you'll end up with multiple heads

nehjoshi:
Since `3.8` has `0030` as the head, we can set the `down_revision` of the new migration (`0033`) to `0030`. Then, we create `0034` with `down_revision: tuple[str, str] = ("0032", "0033")`. 0034 is just for master

nehjoshi:
Now `0033` can be transferred to `3.8`

r00ta (dismissed):

nehjoshi:
Created `0034` and `0035` due to the rebase. `0034` will be transferred to the `3.8` branch. See [PR#510](https://github.com/canonical/maas/pull/510).

r00ta (approved):

PR #506, opened 2026-07-23
docs: MAAS as a snap only docs

r00ta:
Update the documentation removing all the deb package references. MAAS 4.0 is released as snap-only.


MaikRe (changes requested):
Line 288  from `manage-switches.md` needs to be removed as well (I should have left it out in the original PR but did not consider the drop of the deb package)

MaikRe (dismissed):

AloizioMacedo (approved):

PR #510, opened 2026-07-24
feat: add config column to OIDCProviderTable

nehjoshi:
Copied over from [PR#505](https://github.com/canonical/maas/pull/505)
- This PR adds a `config` column to the `OIDCProviderTable`, which will be used to provide any additional config parameters for setting up the provider. For e.g., since we have requests to support Google as an IdP, we require additional attributes (such as `client_email`, `private_key` etc.) for the provider. These attributes can differ between providers, thus the column is of the type `JSONB`. Some providers that are already supported (e.g. Entra ID) do not require additional config params, and thus, this column is nullable.
- The goal is that these additional config options will be shown to the user in the UI/API, and from there we can parse and store this information in an appropriate way in the table.

### DIFF
--- a/src/maasservicelayer/builders/external_auth.py
+++ b/src/maasservicelayer/builders/external_auth.py
@@ -2,6 +2,7 @@
 # GNU Affero General Public License version 3 (see the file LICENSE).
 
 from datetime import datetime
+from typing import Any
 
 from pydantic import Field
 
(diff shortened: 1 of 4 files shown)


AloizioMacedo:
Please don't merge this in 3.8 yet, as we are going through several changes related to the test runs.

nehjoshi:
Sure, marking as `Don't merge` for now

PR #516, opened 2026-07-27
fix: do not retry transactions after losing the database connection

alexsander-souza:
When `@transactional` retried after a retriable failure, `_reset()` reconnected to the database. But a new connectino does not hold the session level advisory lock acquired by `@synchronized` when it's used. This way, the retried code runs outside the critical section entirely unguarded.

On exit, `pg_advisory_unlock` fails on the new connection because it never held the lock, raising `DatabaseLockNotHeld` and crashing. In particular, this was a problem in the rack registration RPC handler.

Remove the reconnection logic so that a broken connection simply fails the retry rather than silently running without the lock.

Resolves LP:2156012

(cherry picked from commit 10cd9a32ac875e0afda150c4fc07522fde8791be)


skatsaounis:
@r00ta based on your comment here: https://bugs.launchpad.net/maas/+bug/2156012/comments/3 I am confused as of what we have to do 1) with this PR 2) the status on the bug report for series affected, etc. Could you please assist?

r00ta:
> @r00ta based on your comment here: https://bugs.launchpad.net/maas/+bug/2156012/comments/3 I am confused as of what we have to do 1) with this PR 2) the status on the bug report for series affected, etc. Could you please assist?

I was wrong when I triaged/fixed the bug, as it really affects also 3.7. I updated the bug and I'm going to merge this

r00ta (approved):

PR #517, opened 2026-07-28
fix: properly change boot order on s390x machines

skatsaounis:
## Summary

Boot-order for non-deploy power actions (on/off/cycle) flowed through the legacy region→rack AMP RPC — a fragile path that duplicated the Temporal boot-order logic already used during deploy. This PR consolidates boot-order onto the Temporal power path, removes the legacy RPC path, switches the --order wire format to JSON across region → agent → maas.power CLI → driver, and fixes boot-disk selection in the deploy workflow.

## Motivation

Boot-order flowed through two divergent paths: the legacy RPC for non-deploy power ops, and the Temporal `DeployWorkflow` during deploy. The RPC path was fragile and duplicated logic. On top of that, the comma-split `--order` format couldn't carry device dicts, and the deploy workflow's boot-disk query joined the wrong table and selected the wrong columns, raising `KeyError: 'serial'`.

## Changes

**Boot-order on the Temporal path**

- Non-deploy power actions (on/off/cycle) now apply boot order through Temporal instead of the region RPC: `PowerParam` carries an optional `boot_order`, and the on/off/cycle workflows apply it via a shared `set-boot-order` agent activity before the power action. `convert_power_action_to_power_workflow()` threads the boot order onto those params. `SetBootOrderParam`/`SET_BOOT_ORDER_ACTIVITY_NAME` moved to `maascommon/workflows/power.py` (shared, avoids a circular import).
- Removed the now-obsolete legacy path: the `SetBootOrder` AMP command (`rpc/cluster.py`) and its rack responder (`rpc/clusterservice.py`), `clusterrpc.power.set_boot_order` and its use in `node._power_control_node`, the orphaned `Node.set_boot_order()` model method, and the `node.set_boot_order()` netboot on/off calls in `metadataserver/api.py`.

**JSON `--order` wire format**

- `provisioningserver/power_driver_command.py`: registers `--order` on the driver subparser and parses it with `json.loads` (was comma-split).
- `maasagent/internal/power/service.go`: forwards boot order as a JSON array, omits `--order` when empty (empty flag is rejected by the CLI), and fixes the `power_params` JSON tag drift.

**Deploy workflow correctness (`maastemporalworker/workflow/deploy.py`)**

- `_get_boot_disk` / `get_boot_order` now `INNER JOIN maasserver_physicalblockdevice` and select the exact fields drivers need (`id`, `name`, `id_path`, `model`, `serial`), fixing `KeyError: 'serial'` and dropping the wrong virtual-block-device join.
- Network boot order is armed before power-on for both ephemeral and regular deploys, matching legacy behavior.

## Out of scope

- HMCZ/IBM Z HMC resilience — the 409,2 "busy" retry and paused-state guard in the HMCZ driver — is handled in a separate follow-up PR.
- The wrongly-reported machine status from the ambiguous paused intermediate state of s390x machines is known, accepted, and intentionally not addressed here.

Resolves: LP:2158480


r00ta (commented):
~~I'm not familiar with s390x so I need more context. In some power drivers the `power_on` and `power_off` operations already set `pxe` as next boot. Should we do the same here?~~ nvm, I changed my mind

skatsaounis:
> ~I'm not familiar with s390x so I need more context. In some power drivers the `power_on` and `power_off` operations already set `pxe` as next boot. Should we do the same here?~ nvm, I changed my mind

My fault to initially link it with the sibling bug (https://bugs.launchpad.net/maas/+bug/2158318), instead of the correct one [2158480](https://bugs.launchpad.net/maas/+bug/2158480), might have contributed to some extra confusion. Even the branch name is wrong here. I apologize. In any case, I fixed the footer.

As a general comment, `hmcz` is the only power driver that we have to set the boot order explicitly. The architecture `s390x` is related to the LPAR (Logical PARtitions), a type of VMs, that a user can compose on an IBM Z Mainframe. HMCZ is a variance of HMC (Hardware Management Console) for the Z Mainframes. IIUC the information available on the web and our source code, the way we do it is the only option.

alexsander-souza (dismissed):
+1

I would try to split this into two PRs: one for the boot order changes and another for the HMCZ-specific improvements

alexsander-souza (approved):
+1

PR #524, opened 2026-07-30
chore: remove RBAC/Candid support

r00ta:
The legacy macaroon-based external authentication (Candid + Canonical RBAC) is removed in favor of the built-in ReBAC authorization (OpenFGA) and OIDC authentication introduced in MAAS 3.8.

This drops the `maas configauth` command, macaroon authentication on the legacy and v3 APIs, the RBAC sync service and triggers, and the macaroonbakery/pymacaroons dependencies. 

Upgrade safety is enforced by two migrations: 
- 0038 aborts the upgrade if external authentication is still configured, so administrators can switch to a local admin account first. Otherwise it permanently deletes Candid/RBAC user accounts (non-local users without an OIDC provider) along with their keys, tokens, notifications and files. Users still owning machines, IP ranges or static IPs block the upgrade until resources are reassigned; operation history is kept with the user reference cleared. 
- 0039 drops the maasserver_rbacsync, maasserver_rbaclastsync and maasserver_rootkey tables, the RBAC sync triggers/procedures and maasserver_userprofile.auth_last_check column. 

BREAKING CHANGE: Candid/RBAC external authentication is no longer supported. Deployments using it must disable external authentication and create a local administrator account before upgrading. The `maas configauth` command is removed.


r00ta:
QA tests manually executed: setup an environment (with and without vault integration) configured with RBAC and Candid. Assign machines, devices, ip addresses and ip ranges to Candid/RBAC users and try to upgrade. The migration successfully abort the upgrade if external_auth is configured or any external user still owns any resource. Once external_auth is disabled and the resources are moved to other local users, the upgrade goes smoothly and MAAS is fully operational.

A migration tool to inspect the environment and re-assign resources can be found here https://gist.github.com/r00ta/724cc68fd88b22a044d6620e1d602afb . It has been used to migrate the test environment. It will be reviewed again and published as part of the official docs

src/maasservicelayer/db/alembic/versions/0038_block_upgrade_if_external_auth_configured.py
```diff
@@ -0,0 +1,200 @@
...
+    # guaranteed to be empty by the check above.
+    dependent_deletes = (
+        "DELETE FROM piston3_token WHERE user_id IN :ids",
+        "DELETE FROM piston3_consumer WHERE user_id IN :ids",
+        "DELETE FROM maasserver_filestorage WHERE owner_id IN :ids",
+        "DELETE FROM maasserver_sshkey WHERE user_id IN :ids",
+        "DELETE FROM maasserver_sslkey WHERE user_id IN :ids",
+        "DELETE FROM maasserver_notification WHERE user_id IN :ids",
+        "DELETE FROM maasserver_oidcrevokedtoken WHERE user_email IN "
+        "(SELECT username FROM auth_user WHERE id IN :ids)",
```
alemar99: on `maasserver_oidcrevokedtoken` we have `ondelete="CASCADE"` so this is not necessary

src/maasservicelayer/db/alembic/versions/0038_block_upgrade_if_external_auth_configured.py
```diff
@@ -0,0 +1,200 @@
...
+        "DELETE FROM auth_user_groups WHERE user_id IN :ids",
+        "DELETE FROM auth_user_user_permissions WHERE user_id IN :ids",
+        # Operation history is kept, but the user reference is cleared (the
+        # column is nullable).
+        "UPDATE maasserver_operation SET user_id = NULL WHERE user_id IN :ids",
```
alemar99, line 176: I think we might want to do the same with `maasserver_event`

src/maasservicelayer/db/alembic/versions/0038_block_upgrade_if_external_auth_configured.py
```diff
@@ -0,0 +1,200 @@
...
+    )
+    for statement in dependent_deletes:
+        conn.execute(
+            text(statement).bindparams(bindparam("ids", expanding=True)),
+            {"ids": user_ids},
+        )
+
+    conn.execute(
+        text(
+            "DELETE FROM maasserver_userprofile WHERE user_id IN :ids"
```
alemar99: We can move this inside `dependent_deletes`

src/maasservicelayer/db/alembic/versions/0039_drop_rbac_and_rootkey_tables.py
```diff
@@ -0,0 +1,88 @@
...
+    )
+
+    # Drop the RBAC sync procedures.
+    op.execute("DROP FUNCTION IF EXISTS sys_rbac_rpool_insert() CASCADE;")
+    op.execute("DROP FUNCTION IF EXISTS sys_rbac_rpool_update() CASCADE;")
+    op.execute("DROP FUNCTION IF EXISTS sys_rbac_rpool_delete() CASCADE;")
+    op.execute("DROP FUNCTION IF EXISTS sys_rbac_config_insert() CASCADE;")
+    op.execute("DROP FUNCTION IF EXISTS sys_rbac_config_update() CASCADE;")
+    op.execute("DROP FUNCTION IF EXISTS sys_rbac_sync_update CASCADE;")
+    op.execute("DROP FUNCTION IF EXISTS sys_rbac_sync() CASCADE;")
```
alemar99: nit: I would put the name without parenthesis for all the functions, since it has to be unique in the db

alemar99 (changes requested):
Maas-sive work, can't wait to remove these!

A couple of nits, otherwise looks good

src/maasservicelayer/db/alembic/versions/0038_block_upgrade_if_external_auth_configured.py
r00ta: I populated the `username` column automatically and then cleaned the user_id

src/maasservicelayer/db/alembic/versions/0038_block_upgrade_if_external_auth_configured.py
```diff
@@ -0,0 +1,198 @@
...
+EXTERNAL_AUTH_CONFIGURED_MESSAGE = (
+    "This MAAS version has removed support for Candid/RBAC external "
+    "authentication, but this deployment still relies on it.\n"
+    "Upgrading now would disable external logins and could lock out "
+    "administrators.\n\n"
+    "Before upgrading, on your current MAAS release:\n"
+    "  1. Make sure a local administrator account with a usable password "
+    "exists (create one with `maas createadmin` if needed).\n"
+    "  2. Disable external authentication by running `maas configauth` and "
+    "leaving the RBAC URL and Candid agent file blank.\n"
```
alemar99: Unfortunately this doesn't fix the issue as there is still the db row:
```
maasdb=# select * from maasserver_secret where path = 'global/external-auth';
            created            |            updated            |         path         |                                        value                                        
-------------------------------+-------------------------------+----------------------+-------------------------------------------------------------------------------------
 2026-09-15 11:11:36.171744+00 | 2026-09-21 17:13:31.883494+00 | global/external-auth | {"key": "", "url": "", "user": "", "domain": "", "rbac-url": "", "admin-group": ""}
(1 row)
```

src/maasservicelayer/db/alembic/versions/0038_block_upgrade_if_external_auth_configured.py
```diff
@@ -0,0 +1,198 @@
...
+                    "SELECT 1 FROM maasserver_vaultsecret "
+                    "WHERE path = :path AND deleted = false LIMIT 1"
+                ),
+                {"path": EXTERNAL_AUTH_SECRET_PATH},
+            ).scalar()
+        )
+
+    return bool(
+        conn.execute(
+            text("SELECT 1 FROM maasserver_secret WHERE path = :path LIMIT 1"),
```
alemar99:
```suggestion
            text("SELECT 1 FROM maasserver_secret WHERE path = :path AND (value::json->>'url' != '' OR value::json->>'rbac-url' != '') LIMIT 1"),
```

alemar99 (approved):
Upgrade from 3.8 to this branch works

PR #526, opened 2026-07-31
fix: normalize overflow in systemd_interval_to_calendar

r00ta:
Intervals above 59 minutes invalid systemd calendar specifications. For example, `120m` generates `*-*-* *:00/120:00` which is invalid. This commit normalizes the input to a valid systemd format.

Fixes LP:2077276


r00ta:
Need more time to work on a solution, will do after my PTO

PR #527, opened 2026-07-31
fix: correct typing annotation for get_dnsresource_queryset

r00ta:
Fix typing annotation for get_dnsresource_queryset

Fixes LP:2003182


skatsaounis (approved):
LGTM Please take a look at the affected series in LP and mark them accordingly. Most probably we do not want to backport it.

PR #529, opened 2026-08-06
feat: populate hardware profile during commissioning

alemar99:
Added Pydantic models to parse and validate commissioning data coming from the `50-maas-01-commissioning` script. The models maps the LXD API models which are internally used by the `machine-resources` binary (which is run in the commissioning script).
All the models have a default, and `null` value are explicitly removed before parsing: this allows us to use the default whenever the data is not present and to avoid null checking on those attributes.

The commmissioning data is parsed into an `HardwareProfile` using some parsing utilities defined in the `lxd_parsing` module. They are derived from what we already do in `src/metadataserver/builtin_scripts/hooks.py`.

I'm still not sure about the `hardware_fingerprint` attribute of the `HardwareProfile`, hence why I left a `FIXME` comment.

The `HardwareProfileBuilder` has been manually tested with some real commissioning data. Also, some tests from `src/metadataserver/builtin_scripts/hooks.py` has been duplicated to assert that the new method parses data correctly.


src/maascommon/storage.py
```diff
@@ -0,0 +1,82 @@
...
+        ),
+    ],
+    "ip": [
+        re.compile(
+            r"^[\.\-\w:]+-iscsi-(?P<target>[\.\-\w:]+)-(?P<lun>lun-(0x)?[\da-fA-F]+)$"
+        )
+    ],
+}
+
+_DEVICE_PATH_RE = re.compile(
```
motrilense: Considering that the value of `_DEVICE_PATH_RE` is generated here and it might not be completely clear what's coming out of re.compile, you could specify it:
```python
_DEVICE_PATH_RE: re.Pattern[str] = re.compile
```

src/maasservicelayer/utils/lxd_parsing.py
```diff
@@ -0,0 +1,255 @@
...
+CPU_MODEL_SPEED_RE = re.compile(r"\s@\s(?P<ghz>\d+\.\d+)GHz$")
+
+
+def parse_cpu(cpu: LXDCPU) -> tuple[int, int]:
+    """Return the CPU core count and speed (MHz) from LXD CPU resources."""
+    socket_names = [socket.name for socket in cpu.sockets if socket.name]
+
+    cpu_speed = 0
+    # Only trust the model name speed when all sockets are the same model.
+    if socket_names and all(name == socket_names[0] for name in socket_names):
```
motrilense: If you use the walrus operator you can reduce the nested blocks and present to the reader of the code all conditions together in the `if` clause

```python
    cpu_speed = 0
    # Only trust the model name speed when all sockets are the same model.
    if (
        socket_names 
        and all(name == socket_names[0] for name in socket_names)
        and (match := CPU_MODEL_SPEED_RE.search(socket_names[0]))
    ):
        cpu_speed = int(float(match.group("ghz")) * 1000)
```

src/maasservicelayer/utils/lxd_parsing.py
```diff
@@ -0,0 +1,255 @@
...
+                cpu.sockets
+            )
+            if average:
+                cpu_speed = round(average / 100) * 100
+
+    return cpu.total, cpu_speed
+
+
+def parse_memory_mb(memory: LXDMemory) -> int:
+    return int(memory.total / 1024**2)
```
motrilense: I usually use bitwise shifts for multiplication and division by power of 2. In this case:
`memory_mib = memory.total >> 20`

But I understand, it might be confusing the 20 there. However, a floor division is a good trade-off:
```python
memory_mib = memory.total // (1024**2)
```

src/maasservicelayer/utils/lxd_parsing.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+def interface_speed(port: LXDNetworkCardPort) -> int:
+    """Return the highest speed (Mbit/s) the port supports."""
+    if not port.supported_modes:
+        return 0
+    return max(int(mode.split("base")[0]) for mode in port.supported_modes)
```
motrilense, line 85: Using 0 as the default max value, you can combined both returns in one expression:
```python
return max((int(mode.split("base")[0]) for mode in port.supported_modes), default=0)
```

src/maasservicelayer/utils/lxd_parsing.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+def disk_id_path(device_id: str, serial: str, disk_id: str) -> str:
+    """Return a stable device path, preferring the by-id link."""
+    id_path = f"/dev/disk/by-id/{device_id}" if device_id else ""
```
motrilense: The reading flow is a bit easier when using positive clauses :)
```python
if device_id and serial:
    id_path = f"/dev/disk/by-id/{device_id}"
else:
    # No by-id link or no serial is a strong indicator of a virtual disk, so
    # fall back to the plain device path.
    id_path = f"/dev/{disk_id}"
```

src/maasservicelayer/utils/lxd_parsing.py
```diff
@@ -0,0 +1,255 @@
...
+
+def parse_network(network: LXDNetwork) -> list[HardwareNetworkGroup]:
+    groups: dict[tuple[str, str, str, str, int], list[HardwareNetworkItem]] = (
+        defaultdict(list)
+    )
+    for card in network.cards:
+        sriov_max_vf = card.sriov.maximum_vfs if card.sriov else 0
+        for port in card.ports:
+            mac = port.address
+            if not mac or len(mac) > MAX_MAC_ADDRESS_LENGTH:
```
motrilense, line 185: out of curiosity, what does having a MAC length greater than the maximum allowed mean? that the network is not valid?

motrilense (commented):
I left a few points

src/maasservicelayer/utils/lxd_parsing.py
alemar99: I think that having the if before makes it a little bit clearer to the reader

src/maasservicelayer/utils/lxd_parsing.py
alemar99: It's for infiniband devices. There is a comment above the constant definition

motrilense (approved):
looks good on my end 🚢

PR #537, opened 2026-08-11
fix: check for indication of virtual device (#512)

AloizioMacedo:
bcache is being interpreted as a physical device when it should not.

~~LXD has a specific extension that can be used to check for usage by any virtual parent devices. It is documented here: https://canonical.com/lxd/docs/default/api-extensions/#resources-disk-used-by~~

~~Note that the fix works in general by checking the presence of something in that key (as documented above). bcache is only an example, even though it was the one that bit us in the test runs.~~

When testing locally (setting up a VM with bcache), the documented field above never came up. According to this comment,
https://github.com/canonical/lxd/blob/ee58e024857aa16e73e4b5e88cdbb55a3fbd5165/lxd/resources/storage.go#L196, it makes sense that it would not be populated for bcache (due to the `pathIsDir` check). But then the use of that key is not clear to me.

The only way I see right now to identify this is to rely on the kernel convention for their names, which start with "bcache". This seems a bit flimsy, but I have no better solution, and this seems to be reliable since the naming is handled directly by the kernel and not something that could be user-specified.

(cherry picked from commit 9d021c24297f26b36701144e04e466ee2874ae41)

### DIFF
--- a/src/metadataserver/builtin_scripts/hooks.py
+++ b/src/metadataserver/builtin_scripts/hooks.py
@@ -810,6 +810,18 @@ _DEV_PATH = re.compile(
 )
 
 
+# bcache virtual holder devices are always named "bcacheN" by the kernel
+# (e.g. /dev/bcache0). This name comes from the "id" LXD reports for a
+# disk, which is just the /sys/class/block/<name> directory name assigned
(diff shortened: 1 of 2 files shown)


tanzin8r (approved):
+1

PR #538, opened 2026-08-11
feat(ci): nosanity input for TiCS manual dispatch

abuyukyi101198:
## Done

- Added workflow_dispatch input for manual nosanity run trigger for the TiCS action


AloizioMacedo (approved):
Thanks, @abuyukyi101198 !

PR #542, opened 2026-08-13
fix: allow handling disks with same serial

AloizioMacedo:
LXD can report the same serial to different physical disks, resulting in problems later since the serial is part of what identifies the disk and if used as an identifier, curtin assumes uniqueness.

It seems that what LXD reports as "serial" depends on its version, and sometimes it might use something that is unique or not, depending on some hardware quirks. This behavior was already similarly observed in https://bugs.launchpad.net/maas/+bug/2102135.

This PR proposes a way to disambiguate the disks similarly to what we already did when we were not able to identify them by model/serial (in this previous case, by the virtue of them perhaps not existing). So essentially there is no deep new logic per se, but rather we add a new case for the disambiguation logic (and a more appropriate error message in case the conflict exists and is not solveable).

### DIFF
--- a/src/maasserver/preseed_storage.py
+++ b/src/maasserver/preseed_storage.py
@@ -785,14 +785,16 @@ def get_disks_needing_path_fallback(
     same model/serial, curtin cannot tell them apart: both storage-config
     disk IDs resolve to the same underlying device, silently producing
     an overlapping storage graph (e.g. two bcache backing devices
-    stacked on one another) instead of a clear failure. This can happen
-    when the hardware/controller firmware reports a duplicate serial
-    across distinct LUNs (as opposed to true multipath, where the same
-    LUN is reachable via more than one path).
+    stacked on one another) instead of a clear failure.
+
+    The above can happen when the hardware/controller firmware reports
+    a duplicate serial across distinct LUNs (as opposed to true
+    multipath, where the same LUN is reachable via more than one path).
+    This was observed, for example, for the case of HPE GEN9 machines
+    using disks in pass-through mode.
 
     For any disk affected by such a collision, curtin can still
-    disambiguate it unambiguously via its `path` key (which resolves
-    via `os.path.realpath()`, bypassing serial/wwn lookup entirely), as
+    disambiguate it unambiguously via its `path` key, as
     long as its `id_path` is set and is itself unique among the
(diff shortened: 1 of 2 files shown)


src/maasserver/preseed_storage.py
```diff
@@ -759,3 +772,54 @@ def compose_curtin_storage_config(node):
...
+    Curtin resolves a disk stanza's identity via `wwn`, then `serial`,
+    by scanning `/dev/disk/by-id/` for a matching entry and picking the
+    shortest match. If two distinct `PhysicalBlockDevice`s report the
+    same model/serial, curtin cannot tell them apart: both storage-config
+    disk IDs resolve to the same underlying device, silently producing
+    an overlapping storage graph (e.g. two bcache backing devices
+    stacked on one another) instead of a clear failure. This can happen
+    when the hardware/controller firmware reports a duplicate serial
+    across distinct LUNs (as opposed to true multipath, where the same
+    LUN is reachable via more than one path).
```
r00ta: I'd be a bit more specific on this. We've experienced this behavior only on HPE GEN9 machines using disks in pass-through mode.

src/maasserver/preseed_storage.py
```diff
@@ -759,3 +772,54 @@ def compose_curtin_storage_config(node):
...
+
+    For any disk affected by such a collision, curtin can still
+    disambiguate it unambiguously via its `path` key (which resolves
+    via `os.path.realpath()`, bypassing serial/wwn lookup entirely), as
+    long as its `id_path` is set and is itself unique among the
+    colliding set. The ids of such disks are returned by this function.
```
r00ta, line 799: This is an implementation detail of curtin that we should not leak into our codebase. The important bit is that we can just provide other identifiers for disks

r00ta (commented):
LGTM, just some nitpicking around the comments

src/maasserver/preseed_storage.py
AloizioMacedo: A little bit difficult for this not to leak a little bit, since we are relying on it being a bona fide identifier for curtin. I will supress the `os.path.realpath` part of the comment, but will refer to curtin still.

r00ta (dismissed):
lgtm

r00ta (approved):

PR #548, opened 2026-08-18
fix: Ensure SSL compatibility with FIPS mode

alexsander-souza:
Updates the HTTP service configuration and Nginx template to conditionally set SSL parameters. When FIPS mode is enabled, it enforces specific curves and ciphers, excluding non-compliant options like X25519 and ChaCha20.


MaikRe (approved):
+1

PR #551, opened 2026-08-19
refactor: fix database connection parameters

MaikRe:
Passing ssl parameters as query parameters does not work for sqlalchemy. Instead these parameters should be passed as part of the connection context.


alexsander-souza (approved):
+1

PR #555, opened 2026-08-20
fix: ensure unconfigured links are not used as default gateways

nehjoshi:
[LP#2162993](https://bugs.launchpad.net/maas/+bug/2162993) states that when an interface is set to "Unconfigured" (i.e. is not configured for an IP address), it still continues to act as the default gateway for a machine. Since the interface is unconfigured, this prevents the machine from receiving outbound network connectivity. Ideally, if there is another active and configured interface on a subnet with a valid gateway, MAAS should point the default_gateways reference to that interface.
- This change can be made by excluding links with no IPs or are unconfigured when getting the default_gateways

Resolves [LP:2162993](https://bugs.launchpad.net/maas/+bug/2162993)


alexsander-souza (approved):
+1

PR #556, opened 2026-08-20
fix(dhcp): restart DHCP on node release and allow OMAPI to update host lease

alemar99:
When we release a node, we set its auto IP to None and then call configure-dhcp:
  a. if it's not a full reload (which isn't if we only have to update the IP) we update DHCP through OMAPI
  b. The OMAPI workflow activity ends in a crash loop because the IP is set to None but the workflow expects a real IP.

Example log:
```
Aug 18 14:43:40 infra1 maas-agent[6264]: ERR Activity error. ActivityType=apply-dhcp-config-via-omapi Attempt=1 Error="unable to decode the activity function input payload with error: payload item 0: unable to decode: invalid IP address: None for function name: apply-dhcp-config-via-omapi" Namespace=default RunID=ef382fab-44b6-4fea-b4af-9beed13fac3e TaskQueue=qsmrdh@agent:main WorkerID=qsmrdh@agent:6264 WorkflowID=configure-dhcp:qsmrdh
```

When running the OMAPI workflow activity, we ignore the error "HostAlreadyExists". This can lead to inconsistencies, i.e. we call AddHost with a new IP-MAC binding but this is ignored. Example log:
```
Aug 18 14:44:32 infra1 maas-agent[6264]: WRN Ignoring already existing host: 52:54:56:58:04:01 ActivityID=4 ActivityType=apply-dhcp-config-via-omapi Attempt=1 Namespace=default RunID=67659744-2b28-4ed3-b3cc-961a0bd0b8bb TaskQueue=qsmrdh@agent:main WorkerID=qsmrdh@agent:6264 WorkflowID=configure-dhcp:qsmrdh WorkflowType=configure-dhcp-for-agent
```

This commit fixes the two bugs above by:
- Passing the subnet_id to configure_dhcp workflow when we set the IP to None
- Deleting and re-adding the host when OMAPI returns HostAlreadyExists

Resolves: LP:2164644

### DIFF
--- a/src/maasagent/internal/dhcp/service.go
+++ b/src/maasagent/internal/dhcp/service.go
@@ -483,46 +483,18 @@ func (s *DHCPService) configureViaOMAPI(ctx context.Context, param ApplyConfigVi
 				return ErrV4NotActive
 			}
 
-			err = clientV4.AddHost(host.IP, host.MAC)
+			err = clientV4.SyncHost(host.IP, host.MAC)
 			if err != nil {
-				if !errors.Is(err, omapi.ErrHostAlreadyExists) {
-					return err
-				}
-
-				// Delete and add the host again if it already exists,
-				// since the binding could have been changed.
-				err = clientV4.DeleteHost(host.MAC)
-				if err != nil {
-					return err
-				}
-
-				err = clientV4.AddHost(host.IP, host.MAC)
-				if err != nil {
-					return err
-				}
+				return err
 			}
 		} else {
 			if !runningV6 {
 				return ErrV6NotActive
 			}
 
-			err = clientV6.AddHost(host.IP, host.MAC)
+			err = clientV6.SyncHost(host.IP, host.MAC)
 			if err != nil {
-				if !errors.Is(err, omapi.ErrHostAlreadyExists) {
-					return err
-				}
-
-				// Delete and add the host again if it already exists,
-				// since the binding could have been changed.
-				err = clientV6.DeleteHost(host.MAC)
-				if err != nil {
-					return err
-				}
-
-				err = clientV6.AddHost(host.IP, host.MAC)
-				if err != nil {
-					return err
-				}
+				return err
 			}
 		}
 	}
--- a/src/maasagent/internal/dhcp/service.go
+++ b/src/maasagent/internal/dhcp/service.go
@@ -489,7 +489,17 @@ func (s *DHCPService) configureViaOMAPI(ctx context.Context, param ApplyConfigVi
 					return err
 				}
 
-				log.Warn(fmt.Sprintf("Ignoring already existing host: %s", host.MAC))
+				// Delete and add the host again if it already exists,
+				// since the binding could have been changed.
+				err = clientV4.DeleteHost(host.MAC)
+				if err != nil {
+					return err
+				}
+
+				err = clientV4.AddHost(host.IP, host.MAC)
+				if err != nil {
+					return err
+				}
 			}
 		} else {
 			if !runningV6 {
@@ -502,7 +512,17 @@ func (s *DHCPService) configureViaOMAPI(ctx context.Context, param ApplyConfigVi
 					return err
 				}
 
-				log.Warn(fmt.Sprintf("Ignoring already existing host: %s", host.MAC))
+				// Delete and add the host again if it already exists,
+				// since the binding could have been changed.
+				err = clientV6.DeleteHost(host.MAC)
+				if err != nil {
+					return err
+				}
+
+				err = clientV6.AddHost(host.IP, host.MAC)
+				if err != nil {
+					return err
+				}
 			}
 		}
 	}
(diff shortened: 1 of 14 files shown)


src/maastemporalworker/workflow/dhcp.py
```diff
@@ -462,24 +459,29 @@ async def _get_hosts_for_reserved_ips(
         )
         result = await tx.execute(stmt)
         return [
-            Host(ip=str(r[0]), mac=str(r[1]) if r[1] else None, hostname="")
-            for r in result.all()
+            Host(
+                ip=str(row["ip"]),
+                mac=str(row["mac_address"]),
+                hostname="",
```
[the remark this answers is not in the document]
alemar99: reserved IP MAC address can't be null

src/maascommon/workflows/dhcp.py
```diff
@@ -43,5 +43,5 @@ def ensure_list(val: list[Any] | None) -> list[Any]:
         ip_range_ids=ensure_list(old.ip_range_ids)
         + ensure_list(new.ip_range_ids),
         reserved_ip_ids=ensure_list(old.reserved_ip_ids)
-        + ensure_list(new.ip_range_ids),
+        + ensure_list(new.reserved_ip_ids),
```
skatsaounis, line 46: IIUC this one-liner is a hidden fix for another (non-recorded) bug. Is my understanding correct?

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -370,20 +370,19 @@ async def find_agents_for_updates(
 
             return AgentsForUpdateResult(agent_system_ids=list(system_ids))
 
-    async def _get_hosts_for_static_ip_addresses(
-        self,
-        tx: AsyncConnection,
-        system_id: str,
-        static_ip_addr_ids: list[int],
-    ) -> list[Host]:
+    async def _get_rack_id(self, tx: AsyncConnection, system_id: str) -> int:
```
skatsaounis, line 373: Does moving this to a separate function contribute to the bug fix or it is a stylistic improvement?

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -830,26 +832,28 @@ async def run(self, param: ConfigureDHCPForAgentParam) -> None:
             # TODO call get_active_interfaces_for_agent and set config
             # directly on the agent
         else:
-            hosts = await workflow.execute_activity(
+            hosts_for_update = await workflow.execute_activity(
```
skatsaounis: nit: I wouldn't rename the variable as part of a bug fix. But it is also fine to do it.

src/maastemporalworker/workflow/dhcp.py
```diff
@@ -830,26 +832,28 @@ async def run(self, param: ConfigureDHCPForAgentParam) -> None:
...
                 FETCH_HOSTS_FOR_UPDATE_ACTIVITY_NAME,
                 FetchHostsForUpdateParam(
                     system_id=param.system_id,
                     static_ip_addr_ids=param.static_ip_addr_ids,
                     reserved_ip_ids=param.reserved_ip_ids,
                 ),
                 start_to_close_timeout=FETCH_HOSTS_FOR_UPDATE_TIMEOUT,
+                result_type=HostsForUpdateResult,
```
skatsaounis: nit: these changes could be omitted from a bug fix PR, since they are not contributing to the fix IIUC. It is also fine to keep them. This line and 849 below.

src/maastemporalworker/workflow/dhcp.py
alemar99: Good catch, was a leftover from another fix, will revert it.

src/maastemporalworker/workflow/dhcp.py
alemar99: Both performance and stylistic. We avoid during the same query twice

src/maascommon/workflows/dhcp.py
alemar99: Yes, that's right

src/maastemporalworker/workflow/dhcp.py
alemar99: Helps with readability, but I can revert them and add to a subsequent PR

src/maasagent/internal/dhcp/service.go
```diff
@@ -489,7 +489,17 @@ func (s *DHCPService) configureViaOMAPI(ctx context.Context, param ApplyConfigVi
...
+				// since the binding could have been changed.
+				err = clientV4.DeleteHost(host.MAC)
+				if err != nil {
+					return err
+				}
+
+				err = clientV4.AddHost(host.IP, host.MAC)
+				if err != nil {
+					return err
+				}
```
AloizioMacedo: There is a little bit of duplication here that can be avoided. Might be worth creating a function

```go
// Adds host, deleting and adding the host again if it already exists,
// since the binding could have been changed.
func syncHost(client omapi.OMAPI, host Host) error {
	err := client.AddHost(host.IP, host.MAC)
	if err != nil {
		if !errors.Is(err, omapi.ErrHostAlreadyExists) {
			return err
		}

		err = client.DeleteHost(host.MAC)
		if err != nil {
			return err
		}

		err = client.AddHost(host.IP, host.MAC)
		if err != nil {
			return err
		}
	}

	return nil
}
```

and doing `syncHost` with `clientV4` here, and `clientV6` in the other clause.

alemar99: Good point, I'll make the change

AloizioMacedo (approved):

PR #558, opened 2026-08-20
fix: Update temporal service configuration and binding

alexsander-souza:


docs/how-to-guides/enhance-maas-security.md
```diff
@@ -233,8 +233,9 @@ A value of `1` means FIPS mode is active.
 ### Activate hardening on a non-FIPS host
 
 Run `maas config-hardening enable`. This sets `hardening_enabled=on` in the
-MAAS database and seeds `prometheus_bind` and `temporal_bind` to `127.0.0.1`
-in `regiond.conf` if they are not already set:
+MAAS database and seeds `prometheus_bind` to `127.0.0.1` in `regiond.conf`
+if it is not already set. `temporal_bind` is left unset: MAAS derives a
+specific, non-wildcard address for it from `maas_url` at startup.
```
MaikRe, line 238:
```suggestion
MAAS database and seeds `prometheus_bind` to `127.0.0.1` in `regiond.conf`
if it is not already set.
```
The wording here is confusing as it makes it seem like the reader is expecting the enable command to set temporal_bind and that its a new change that this is not the fact

docs/reference/configuration-guides/security-hardening.md
```diff
@@ -42,9 +42,10 @@ subcommands:
   list                List all hardening parameters with values and stores.
   validate            Run hardening validation; print violations; exit
                       non-zero if any exist.
-  enable              Set hardening_enabled=on and seed loopback defaults for
-                      unset region-internal bind addresses (prometheus_bind,
-                      temporal_bind).
+  enable              Set hardening_enabled=on and seed a loopback default
+                      for unset prometheus_bind. temporal_bind is left
+                      unset so MAAS can derive a specific address from
+                      maas_url at startup.
```
MaikRe, line 48: Same as previous comment. I would either mention all things that are not set by the enable command (rpc bind api bind etc) and why those are not set or only mention prometheus_bind

docs/reference/configuration-guides/security-hardening.md
```diff
@@ -60,8 +61,10 @@ are written to `regiond.conf` on the local host. The `set` command handles
 YAML quoting automatically.
 
 `enable` is a convenience shortcut: it sets `hardening_enabled=on` and also
-seeds `prometheus_bind` and `temporal_bind` to `127.0.0.1` in `regiond.conf`
-if those keys are unset, saving a separate step on first activation.
+seeds `prometheus_bind` to `127.0.0.1` in `regiond.conf` if unset, saving a
+separate step on first activation. `temporal_bind` is not seeded: MAAS
```
MaikRe, line 65: ""

src/maasserver/management/commands/config_hardening.py
```diff
@@ -30,8 +30,12 @@
     }
 )
 
-# Seeded to loopback by `enable`; network-facing binds are not auto-seeded.
-_LOOPBACK_SEED_KEYS = frozenset({"prometheus_bind", "temporal_bind"})
+# Seeded to loopback by `enable`; network-facing binds are not
+# auto-seeded. `temporal_bind` is excluded for the same reason it's
+# excluded from `maasservicelayer.services.hardening._AUTO_DERIVED_BIND_KEYS`:
+# it's derived from `maas_url` at runtime when unset, so seeding it here
+# would only pin a stale value.
```
MaikRe: This comment could potentially confuse future maintainers as it once again only covers temporal.
It might be good to state which values default to what and why in the documentation and leave additional notes like this out

src/maasserver/tests/test_commands_config_hardening.py
```diff
@@ -180,7 +180,9 @@ def test_sets_config_and_seeds_unset_binds(self):
             self._cmd(command="enable")
 
         self.assertEqual("127.0.0.1", mock_cfg.prometheus_bind)
-        self.assertEqual("127.0.0.1", mock_cfg.temporal_bind)
+        # temporal_bind is not seeded: RegionTemporalService derives a
+        # specific address from maas_url at runtime when left unset.
+        self.assertEqual("", mock_cfg.temporal_bind)
```
MaikRe:
```suggestion

```

src/maastemporalworker/worker.py
```diff
@@ -17,13 +17,14 @@
 from temporalio.service import RPCError, RPCStatusCode
 from temporalio.worker import Worker as TemporalWorker
 
+from maasserver.config import get_temporal_connect_address
```
MaikRe, line 20: This might be better placed in maascommon or somewhere similar in order to avoid the cross dependency between the temporalworker and the maasserver region

src/tests/maasserver/regiondservices/test_temporal.py
```diff
@@ -75,30 +102,30 @@ def _run_configure(
...
 
-    return tpl.environs[0]
+    return tpl.environs[0], mock_get_source_address.call_count
 
 
 @pytest.mark.parametrize("sslmode", ["prefer", "allow", "disable"])
 def test_tls_disabled_for_non_encrypting_modes(sslmode, tmp_path, monkeypatch):
     monkeypatch.setenv("MAAS_TEMPORAL_CONFIG_DIR", str(tmp_path))
-    environ = _run_configure(_db_with_sslmode(sslmode), monkeypatch)
+    environ, _ = _run_configure(_db_with_sslmode(sslmode), monkeypatch)
```
MaikRe, line 121: Style choice but I am personally not a fan of the silent discards here. Would it make sense to make the return type a small dataclass or something similar?

docs/reference/configuration-guides/security-hardening.md
```diff
@@ -72,7 +75,8 @@ if those keys are unset, saving a separate step on first activation.
 | `api_bind` | `regiond.conf` (per-host) | empty | IPv4 address the public API binds to. A specific (non-wildcard) address is required when hardening is active. |
 | `api_bind6` | `regiond.conf` (per-host) | empty | IPv6 address the public API binds to. A specific (non-wildcard) address is required when hardening is active. |
 | `prometheus_bind` | `regiond.conf` (per-host) | empty | IPv4 address the Prometheus metrics endpoint binds to. Seeded to `127.0.0.1` by `maas config-hardening enable` if unset. |
-| `temporal_bind` | `regiond.conf` (per-host) | empty | IPv4 address the Temporal worker binds to. Seeded to `127.0.0.1` by `maas config-hardening enable` if unset. |
+| `temporal_bind` | `regiond.conf` (per-host) | empty | IPv4 address the Temporal services bind to. Left unset by default: derived from `maas_url` at startup, on every install mode (region, rack+region, all-in-one). Set explicitly to pin it elsewhere. |
+| `temporal_server` | `rackd.conf` (per-host) | empty | Address MAAS Agent dials to reach Temporal; not a hardening key. Left unset by default: derived from `maas_url` at startup, the same as `temporal_bind`. Set explicitly to pin it elsewhere. |
```
MaikRe, line 79: Since this is specifically not a hardening key it might be nice to place this at the end of the list to make the difference clear

src/maasagent/cmd/maas-agent/main.go
```diff
@@ -155,13 +156,29 @@ func getClusterCert() (tls.Certificate, *x509.CertPool, error) {
 	return cert, ca, nil
 }
 
+// temporalHost returns the address MAAS Agent should dial to reach
+// Temporal, preferring the explicit TemporalServer config over the
+// first region controller endpoint. Returns an empty string if neither
+// is available.
+func temporalHost(cfg *config) string {
```
MaikRe, line 163: If I understand this correctly then if there is no config item this defaults to the maasl_url of the first available region. Is this always guaranteed to be correctly reachable by the agent?
Can it ever happen that this is a proxy of some kind that doesn't forward the temporal traffic?

docs/how-to-guides/enhance-maas-security.md
alexsander-souza: I'll push a MP addressing the docs later on.

src/maasagent/cmd/maas-agent/main.go
alexsander-souza: It's not guaranteed to be reachable, and in those cases the user should explicitly set maas_url.

src/maastemporalworker/worker.py
alexsander-souza: This depends on RegionConfiguration (from `maasserver`), so it cannot be moved to `maascommon`.

PR #563, opened 2026-08-21
Fips 3.7 password policy

MaikRe:
Enabling hardening on a non FIPS host means that the hardening state needs to be read from the database for every password related command (createadmin, changepassword) run in the CLI as otherwise the password policy is not enforced.


alexsander-souza (approved):
+1

PR #564, opened 2026-08-21
fix: do not create git tags during version bump

tanzin8r:
Version bump no longer creates or pushes git tags. We were tagging at the start of the release process, which is too early.

I'll backport this to 3.7. Creating the tag later will be a separate maas-bot change.


wyattrees (approved):
+1

PR #571, opened 2026-08-25
docs: add docs for multiple boot sources

alemar99:


docs/explanation/images.md
```diff
@@ -18,6 +18,20 @@ The candidate stream contains images and bootloaders which have not been explici
 
 This stream is available [here](http://images.maas.io/ephemeral-v3/candidate).
 
+### Multiple image streams
+
+MAAS 3.8 and later support multiple image streams. This allows you to configure multiple SimpleStreams sources simultaneously, each with a configurable priority. When the same boot resource (for example, Ubuntu Jammy for amd64 architecture) is available in multiple streams, MAAS downloads it from the stream with the highest priority.
```
tmerten: Since versions are now handled by RTD, I don't think we need to mention 3.8 and later anymore. If somebody browses the 3.7 docs they should not see this anyways. 
So this could be "MAAS allows you to configure..."

WDYT @AloizioMacedo ?

docs/explanation/images.md
```diff
@@ -18,6 +18,20 @@ The candidate stream contains images and bootloaders which have not been explici
...
+
+Multiple streams are useful when you:
+
+- Want to test newer images from the candidate stream while keeping production-ready stable images
+- Operate a custom mirror alongside the official Canonical streams
+- Need flexibility to switch between different image sources based on your deployment requirements
+
+Priority values determine precedence: higher values take priority. As a consequence, two streams can't have the same priority.
+
```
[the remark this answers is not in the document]
tmerten: +1. Also you already said this in the paragraph before. Not sure if that needs to be repeated. Can move the sentence from the previous section to here.

docs/explanation/images.md
```diff
@@ -18,6 +18,20 @@ The candidate stream contains images and bootloaders which have not been explici
...
+- Want to test newer images from the candidate stream while keeping production-ready stable images
+- Operate a custom mirror alongside the official Canonical streams
+- Need flexibility to switch between different image sources based on your deployment requirements
+
+Priority values determine precedence: higher values take precedence. As a consequence, two streams can't have the same priority.
+
+You can also disable a boot source to prevent MAAS from downloading images from it. Disabled sources are ignored during image synchronization and won't be considered when multiple streams provide the same image.
+
 ### The retired daily stream
```
tmerten, line 42: This might be hard to understand: What is a retired daily stream? I think you want to tell the user something about what happens on update here. I'd change the heading to reflect that and possibly add other things that are important to know.

docs/explanation/images.md
```diff
@@ -18,6 +18,20 @@ The candidate stream contains images and bootloaders which have not been explici
...
 
 Previously there was only one MAAS stream available, daily. This stream has been replaced by the stable stream. Any client using this stream will be automatically redirected to the stable stream.
```
tmerten: ... available: `daily`.

replaced by the `stable`...

docs/how-to-guides/manage-images.md
```diff
@@ -27,6 +27,76 @@ CLI
 BOOT_SOURCE_ID=$(maas $PROFILE boot-sources read)
 ```
 
+## Multiple image streams support
+
+Starting from version 3.8, MAAS supports multiple boot image streams. You can configure multiple SimpleStreams sources simultaneously, each with a priority value. When the same image is available in multiple streams, MAAS downloads it from the one with the highest priority.
```
tmerten: same as in first comment

tmerten (changes requested):
some nits inline

docs/explanation/images.md
```diff
@@ -18,6 +18,20 @@ The candidate stream contains images and bootloaders which have not been explici
...
+
+Multiple streams are useful when you:
+
+- Want to test newer images from the candidate stream while keeping production-ready stable images
+- Operate a custom mirror alongside the official Canonical streams
+- Need flexibility to switch between different image sources based on your deployment requirements
+
+Priority values determine precedence: higher values take precedence. As a consequence, two streams can't have the same priority.
+
+You can also disable a boot source to prevent MAAS from downloading images from it. Disabled sources are ignored during image synchronization and won't be considered when multiple streams provide the same image.
```
tobiasdemendonca: Is it worth being explicit about what happens to images that were synced from this (now disabled) boot source here, or is it obvious?

docs/explanation/images.md
alemar99: What if we delete this section at all?

docs/explanation/images.md
alemar99: Good point, thanks!

tmerten (approved):
ltgm, thanks

PR #579, opened 2026-08-26
refactor: pass power ssh keys to temporal workflow before starting tasks

MaikRe:
Power driver control via SSH is done on a separate thread meaning that SSH keys for the target power controller need to be injected into the temporal workflow at launch time as the RPC calls will otherwise fail on a FIPS system.


src/maasagent/internal/power/service.go
```diff
@@ -301,21 +313,21 @@ func (s *PowerService) PowerReset(ctx context.Context, param PowerResetParam) (*
...
+	PowerParams PowerParam       `json:"power_param"`
 }
 
 func (s *PowerService) SetBootOrder(ctx context.Context, param SetBootOrderParam) error {
 	log := activity.GetLogger(ctx)
 
 	log.Info("setting boot order of " + param.SystemID)
 
-	_, err := powerCommand(ctx, "set-boot-order", false, param.PowerParams.DriverType, param.PowerParams.DriverOpts)
+	_, err := powerCommand(ctx, "set-boot-order", false, param.PowerParams.DriverType, param.PowerParams.DriverOpts, param.PowerParams.TrustedSSHHostKeys)
```
tanzin8r: `param.Order` isn't passed through here. SetBootOrderParam includes an Order field, and the deploy workflow populates it. Since powerCommand now takes trustedKeys before the bootOrder, this should probably include `param.Order`

src/maasagent/internal/power/service_test.go
```diff
@@ -574,3 +583,175 @@ func TestPowerResetDPU(t *testing.T) {
...
+				"power_user":    "maas",
+				"power_pass":    "maas",
+			},
+			DriverType:         "wedge",
+			TrustedSSHHostKeys: trustedKeys,
+		},
+		Order: []map[string]any{{"boot_type": "network", "device": "pxe"}},
+	}
+
+	expectedArgs := []string{"set-boot-order", "wedge", "--power-address", "10.0.0.1", "--power-pass", "maas", "--power-user", "maas", "--order", "''"}
```
tanzin8r: same as above. This test passes a non-empty Order but expects --order ' '. That only seems to be passing because SetBootOrder doesn't forward param.Order. This looks pre-existing.

src/provisioningserver/drivers/power/ssh_utils.py
```diff
@@ -91,42 +95,101 @@ def _lookup_trusted_key(
     ) -> bool:
         """Return True iff the host key is trusted per the MAAS region.
 
+        Tries RPC first (available when running in rackd). Falls back to
```
alexsander-souza: try via env first, it's cheaper

alexsander-souza (changes requested):

alexsander-souza (approved):

PR #582, opened 2026-08-26
fix: allow multiple bind addresses for hardening keys

alexsander-souza:
Hardening bind keys (api_bind, api_bind6, rpc_bind, dns_bind, ...) only
accepted a single address and several keys (internal_api_bind,
syslog_bind, http_proxy_bind[6], dns_bind6) existed in name only,
without being read by any service. rpc_bind additionally fell back to
a hardcoded 127.0.0.1 under hardening, which broke rack connectivity
whenever it was left unset.

Bind keys are now comma-separated lists (regiond.conf/rackd.conf), and
api_bind/api_bind6/rpc_bind/internal_api_bind[6] derive a specific,
non-wildcard address from maas_url when unset, instead of an
unreachable loopback default. dns_bind/dns_bind6 remain snap-only,
since MAAS does not own named.conf.options on Debian-packaged installs.


.gitignore
```diff
@@ -44,3 +44,4 @@ __pycache__
 .#*
 *.*.swp
 /.dev-certificates
+.ai/
```
MaikRe: This is unrelated to the rest of the changes right?

src/maasservicelayer/services/hardening.py
```diff
@@ -241,27 +287,29 @@ def _validate_bindings(self) -> list[HardeningViolation]:
...
-                            f"Run: maas config-hardening set {key} "
-                            f"<specific-ip-address>"
-                        ),
-                        config_key=key,
+            for value in values:
+                try:
+                    addr = ipaddress.ip_address(value)
+                except ValueError:
+                    violations.append(
+                        _violation(
```
MaikRe, line 295: If there are multiple IP addresses that would lead to a value error only the first one is caught so an operator would not notice that issue until the restart MAAS and the next one is flagged

src/provisioningserver/utils/network.py
```diff
@@ -1095,15 +1108,94 @@ def resolve_bind_address(
...
+    maas_url = ""
+    try:
+        with open_config() as config:
+            configured = list(getattr(config, bind_attr))
+            raw_maas_url = getattr(config, maas_url_attr)
+            if isinstance(raw_maas_url, (list, tuple)):
+                raw_maas_url = raw_maas_url[0] if raw_maas_url else ""
+            maas_url = str(raw_maas_url)
+    except Exception:
+        pass
```
MaikRe: Is swallowing all exceptions here safe or should there at least be a log statement to output what went wrong?

snap/local/tree/bin/run-named
```diff
@@ -6,6 +6,10 @@
 mkdir -p "$SNAP_DATA/bind"
 cp -f "$SNAP"/etc/bind/* "$SNAP_DATA/bind/"
 
+# The stock named.conf.options shipped with the bind9 package is not
+# included by named.conf below, so remove it.
+rm -f "$SNAP_DATA/bind/named.conf.options"
```
MaikRe, line 11: Any chance an operator would modify this before a snap restart and then be surprised when it dissapears?

src/maasserver/rpc/regionservice.py
```diff
@@ -942,53 +950,73 @@ def _removeConnectionFor(self, ident, connection):
...
             except Exception:
                 if index == last:
-                    raise
+                    log.err(
+                        Failure(), "RegionServer endpoint failed to listen."
+                    )
             else:
-                returnValue(port)
+                returnValue(ports)
+        returnValue([])
```
MaikRe, line 1009: Rather than raising an error if no ports bind correctly, this now returns an empty list.
Is this case properly handled by the caller of this function?

src/maasserver/ipc.py
MaikRe, line 75: Related to the other comment, if the ports list from _bindFirstGroup is empty this will receive a 'None' from getPort (`src/maasserver/rpc/region.py` line 1060) and may raise a type error

MaikRe (changes requested):
A few comments and some questions

snap/local/tree/bin/run-named
alexsander-souza: the file was not used, so any change to it was innocuous already

MaikRe (approved):
+1

PR #586, opened 2026-08-27
fix: increase rate and burst limit

alexsander-souza:
when using TLS, the browser uses HTTP/2 which in turn allows a higher number of concurrent requests, exceeding the old burst limit


tanzin8r (approved):
+1

PR #589, opened 2026-08-28
fix: exempt Unix socket DB connections from INSECURE_DB_SSLMODE

alexsander-souza:
Skip the database_sslmode hardening check when database_host is a filesystem path (Unix domain socket), e.g. as used by maas-test-db. libpq never negotiates TLS over a Unix socket, so sslmode is not applicable and should not trigger a violation.


MaikRe (approved):
+1, assuming merge conflicts are resolved

PR #590, opened 2026-08-30
fix(hardening): close FIPS/hardening enforcement gaps

alexsander-souza:
Region and rack now share one bind-validation implementation instead of enforcing security policy from two independently maintained copies.

- ipmi: default cipher_suite_id is FIPS-compliant (17) instead of the fail-open default that the FIPS validator itself rejected
- api/users: v2 user creation enforces password complexity under hardening, matching v3 and createadmin
- hardening: resolution strings quote CLI forms the shell actually accepts (`maas config-hardening ...`, `maas-rack config-hardening ...`)
- config: drop dead `hardening_enabled` regiond.conf option; hardening state is DB-backed only
- config-hardening validate: always runs the validator so FIPS drift is reported instead of being skipped
- provisioningserver: add rack hardening CLI (list/get/set/validate) wired into maas-rack, backed by the shared maascommon.hardening bind-violation checker
- hardening: a malformed bind address wins over a wildcard finding for the same key instead of being silently downgraded
- fips: RSA key-size check fails closed on parse error
- forms/fips_power: emit fips_driver_rejected audit event on rejection
- prometheus/config-hardening: stop seeding prometheus_bind; enable reports partial failures instead of a false success message
- services/hardening: delegate _validate_bindings() to maascommon.hardening.check_bind_violations() instead of a duplicated implementation; fixes INVALID_BIND_ADDRESS idents colliding across different bind keys (same notification-collapse class as F8, missed because the region path never called the shared helper)

### DIFF
--- a/src/maasservicelayer/services/hardening.py
+++ b/src/maasservicelayer/services/hardening.py
@@ -279,7 +279,7 @@ class HardeningValidator:
         Delegates to `maascommon.hardening.check_bind_violations`, the same
         implementation used by the rack's `hardening_command` CLI, so the
         region and rack enforce identical rules from a single source of
-        truth (see F8 in the hardening findings).
+        truth.
         """
         return [
             _violation(
--- a/src/maasservicelayer/services/hardening.py
+++ b/src/maasservicelayer/services/hardening.py
@@ -12,7 +12,6 @@ The validator never raises, exits, or blocks socket binding.
 
 from collections.abc import Sequence
 from dataclasses import dataclass, field
-import ipaddress
 import logging
 from pathlib import Path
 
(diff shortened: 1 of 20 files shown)


src/maasservicelayer/services/hardening.py
```diff
@@ -275,59 +274,25 @@ def _validate_dh_params(self) -> list[HardeningViolation]:
...
-            resolution=(
-                f"Run: maas config-hardening set {key} <specific-ip-address>"
-            ),
-            config_key=key,
-            ident=f"hardening-wildcard-bind-{key.replace('_', '-')}",
-        )
+        Delegates to `maascommon.hardening.check_bind_violations`, the same
+        implementation used by the rack's `hardening_command` CLI, so the
+        region and rack enforce identical rules from a single source of
+        truth (see F8 in the hardening findings).
```
MaikRe: This comment lacks context as to what "F8 in the hardening findings" is

alexsander-souza: this leaked from my spec, will remove it

MaikRe (approved):
+1, pending removal of aforementioned comment

PR #597, opened 2026-09-01
fix: enforce ReBAC authorization checks in websocket handlers

r00ta:
Enforce authorization checks in websocket handlers


src/maasserver/auth/local.py
```diff
@@ -37,6 +39,8 @@
     DNSResource,
     Domain,
     Fabric,
+    IPRange,
+    ReservedIP,
```
alemar99, line 43: leftovers?

alemar99 (commented):
Looks good, one leftover

src/maasserver/auth/local.py
alemar99: sorry, misread it

alemar99 (approved):

PR #598, opened 2026-09-01
feat: UI sync smoke test and version matrix

abuyukyi101198:
## Done

- Fixed the workflow bug that would smoke test the to-be-replaced UI commit instead of the incoming commit
- Added cypress test artifact upload on failure
- Skips smoke test and sync if UI AND MAAS are unchanged, runs smoke test if only MAAS is changed, runs smoke test and sync if UI is changed
- Added matrix runs for 3.8, 3.7, and 3.6

### DIFF
--- a/.github/workflows/maas-ui-sync.yaml
+++ b/.github/workflows/maas-ui-sync.yaml
@@ -147,7 +147,10 @@ jobs:
           npx cypress run
 
       - name: Commit submodule update
-        if: steps.compare.outputs.has_update == 'true'
+        if: |
+          steps.compare.outputs.has_update == 'true' &&
+          steps.compare.outputs.should_run == 'true' &&
+          matrix.maas_branch != 'debug'
         env:
           LATEST_COMMIT: ${{ steps.compare.outputs.latest_commit }}
           TARGET_BRANCH: ${{ steps.compare.outputs.target_branch }}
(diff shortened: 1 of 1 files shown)


.github/workflows/maas-ui-sync.yaml
```diff
@@ -11,49 +11,91 @@ permissions:
...
+  sync:
+    name: Sync ${{ matrix.maas_branch }}
     runs-on: ubuntu-latest
-    outputs:
-      has_update: ${{ steps.compare.outputs.has_update }}
-      latest_commit: ${{ steps.compare.outputs.latest_commit }}
+    strategy:
+      fail-fast: false
+      matrix:
+        maas_branch: ${{ fromJSON(github.event_name == 'push' && '["debug"]' || '["master", "3.6", "3.7", "3.8"]') }}
```
ndv99, line 20: Will this match all branches called "debug"? I think this should be a bit more specific

abuyukyi101198: No, that condition sets the matrix entry of any push-triggered run of this workflow as `"debug"`. Since only branches beginning with `"maas-ui-sync-debug-*"` accept on-push runs (rule at the top of the file), it works for any branch that matches `"maas-ui-sync-debug-*"`.

abuyukyi101198: It is a catch-all matrix label for debug branches.

nehjoshi (approved):
LGTM

PR #603, opened 2026-09-02
docs: setup release information

AloizioMacedo:
Sets up the release-information branch, which will act as source of truth for any update and tracking of things such as release notes, supported maas versions etc.


skatsaounis (approved):
Thank you for the clarification to all of my points. Let's make our version agnostic pages true version agnostic. LGTM

PR #605, opened 2026-09-02
fix: properly direct fips_ssh_authentication log statements to MAAS syslog

wyattrees:
This PR fixes audit logging for fips_ssh_authentication events. Previously, info-level log statements made by power drivers would not reach the MAAS syslog.


tanzin8r (approved):
+1

PR #609, opened 2026-09-03
fix: openfga wouldn't start if vault was enabled

bjornt:
openfga always tried to get the database credentials from regiond.conf. But if Vault is used, there are no DB credentials in regiond.conf.

Follow the same pattern as Temporal and write a config file, including the database credentials, before starting openfga.

Resolves LP:2166344

### DIFF
--- a/snap/local/tree/usr/bin/run-openfga
+++ b/snap/local/tree/usr/bin/run-openfga
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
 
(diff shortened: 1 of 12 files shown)


r00ta (changes requested):
Writing the DB credentials to a file bypasses our goal of keeping secrets managed in Vault. What do you think about adopting the approach from https://github.com/canonical/maas/blob/a04d64902b0ed3ed727777c200c2cd4983efaa2d/src/maasapiserver/settings.py#L48 in the main of the OpenFGA wrapper instead?

PR #611, opened 2026-09-04
fix: Add missing fips_tls_handshake logs for remaining outbound TLS connections

wyattrees:
The fips_tls_handshake audit log was missing for outbound TLS connections created for downloading boot resources, Site Manager communication, and vault and macaroon management


alexsander-souza (approved):
+1

PR #619, opened 2026-09-04
feat: allow specific scripts in CSP header for FIPS

MaikRe:


alexsander-souza (changes requested):
we need to split this in 3 PRs:

1. off-line assets should go to `master` first and be back-ported to `3.8` and `3.7` later, and then we rebase `fips-3.7`
2. the v2 docs fix should follow the same process
3. after all back-ports and rebases, only FIPS/Hardening changes should remain in this PR

docs/reference/configuration-guides/security-hardening.md
```diff
@@ -208,3 +208,89 @@ Structured JSON events. View them with `journalctl -o json`.
...
+```
+
+Reads the user's stored theme preference from `localStorage` and applies it to
+`<body>` before the page paints, so the correct theme is used from the first
+frame. Without this script the page briefly flashes in the wrong theme, and
+the light/dark/auto toggle no longer takes effect on the first visit.
+
+Source: [Furo theme](https://github.com/pradyunsg/furo).
+
+### Regenerating a hash
```
tanzin8r: I think this part is more maintainer documentation than operator reference. We already keep similar update procedures next to the assets themselves (for example [docs/_static/fonts/README.md](https://github.com/canonical/maas/blob/fips-3.7/docs/_static/fonts/README.md) and [swagger-ui-dist/README.md](https://github.com/canonical/maas/blob/fips-3.7/src/maasserver/templates/swagger-ui-dist/README.md)).
can we keep a short CSP summary in this guide and move the regeneration steps elsewhere?

alexsander-souza: +1, this is maintainer documentation

alexsander-souza (approved):
Image

PR #625, opened 2026-09-08
fix(ci): maas-ui-sync chrono run for all version branches

abuyukyi101198:
## Done

- maas-ui-sync Cypress version pinned (failure reason)
- maas-ui-sync now runs daily for master, 3.8, 3.7 and 3.6
- maas-ui-sync now runs for all branches when triggered manually or through a debug branch, but debug branch runs do not publish the submodule


ndv99 (approved):
LGTM

PR #626, opened 2026-09-08
Resolve fips conflicts 3.7

MaikRe:
Cherry picks commits from 3.7 not currently in fips-3.7 in order to prepare for merging of fips branch.


alexsander-souza (changes requested):
let's rebase `fips-3.7` on top of a `3.7` branch that already has all these fixes. cherry-picking unrelated changes into a feature branch often makes very hard to port it forward.

PR #633, opened 2026-09-08
fix: hide DB-only regiond commands on rack-only snap

alexsander-souza:
apikey, configauth, config-tls, config-vault, msm, createadmin and changepassword all require the region database. On a rack-only snap there is no local database, so these commands were listed but would fail at runtime. Hide them from the CLI in that mode instead.


alemar99 (approved):
LGTM

PR #635, opened 2026-09-08
feat: add PUT for trusted SSH host keys

tanzin8r:
Adding the missing PUT for trusted SSH host keys so the API matches the spec.

### DIFF
--- a/src/maasservicelayer/services/ssh_host_keys.py
+++ b/src/maasservicelayer/services/ssh_host_keys.py
@@ -31,6 +31,18 @@ class TrustedSshHostKeysService(
         super().__init__(context, repository)
 
     async def pre_create_hook(self, builder: TrustedSshHostKeyBuilder) -> None:
+        self._validate_fips_public_key(builder)
+
+    async def pre_update_instance(
+        self,
+        existing_resource: TrustedSshHostKey,
+        builder: TrustedSshHostKeyBuilder,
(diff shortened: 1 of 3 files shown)


wyattrees (approved):
+1

src/maasservicelayer/services/ssh_host_keys.py
```diff
@@ -31,6 +31,18 @@ def __init__(
         super().__init__(context, repository)
 
     async def pre_create_hook(self, builder: TrustedSshHostKeyBuilder) -> None:
+        self._validate_fips_public_key(builder)
+
+    async def pre_update_instance(
```
MaikRe, line 36: This is technically a misuse of this function as it is intended to be used to act on the instance and not the builder.
I would personally be in favor of adding a pre_update_hook to `services/base.py` matching the current architecture regarding hooks and using that instead.

MaikRe (changes requested):

src/maasservicelayer/services/ssh_host_keys.py
MaikRe: We could also leave it as it is for now and add the pre_update_hook to master first, backport it to 3.8 and 3.7 and then use it here in order to not need to port this forward.

MaikRe (approved):
+1

PR #646, opened 2026-09-10
fix: do not create deployment scripts for diskless deployments

r00ta:
Ephemeral deployments do not use deployment scripts.

Resolves LP#2166890


AloizioMacedo (changes requested):
I think we should implement the first copilot suggestion. Not only due to the cleanup points it mentions, but also since if the function owns the mutation of the script set, it is best for it to own in all conditional cases (not only in case it is not a ephemeral deployment).

AloizioMacedo (approved):

PR #648, opened 2026-09-10
refactor: unify IPv4/IPv6 hardening bind config keys

alexsander-souza:
api_bind/api_bind6, api_int_bind/api_int_bind6, agent_api_bind/ agent_api_bind6, http_proxy_bind/http_proxy_bind6, and dns_bind/ dns_bind6 collapse into single mixed-family keys, matching the rpc_bind/syslog_bind/tftp_bind pattern already in use. Reduces the config-hardening surface from 10 split keys to 5 merged ones.

A new resolve_dual_stack_bind_addresses()/resolve_dual_stack_service_bind() pair in provisioningserver.utils.network preserves today's per-family auto-derivation: pinning one family explicitly still backfills the other family's maas_url-derived default under hardening, instead of the naive "non-empty list wins verbatim" rule silently dropping it.

nginx, squid, and BIND9 renderers now classify addresses by family at render time; api_int_bind gains list support (previously scalar) ahead of its first release. The hardening feature has not shipped in any 3.7.x point release, so no regiond.conf/rackd.conf migration or v3 API deprecation path is required.


MaikRe (approved):
+1

PR #654, opened 2026-09-11
refactor: consolidate FIPS validation and dedupe hardening code

alexsander-souza:
Follow-up cleanup from a code-quality review of the FIPS/hardening branch: all fixes but one are dead-code removal, or filling narrow observability/consistency gaps. the `rpc_bind` change that is a regression fix.

- Consolidate three independent implementations of the FIPS TLS/SSL key-compliance rule (weak signature algorithm, DSA key, RSA <2048 bits) into a single maascommon.fips.validate_fips_tls_certificate(), used by the hardening validator, the SSL key service, and the legacy Django SSL key model (migrated from pyOpenSSL to `cryptography` to share the check). Dedupe the insecure-sslmode constant between maasservicelayer.db and the hardening service into maascommon.hardening.INSECURE_DB_SSLMODES.
- Simplify maascli's rack/region command gating: replace the DB_ONLY_REGIOND_COMMANDS/skip frozenset machinery (expressing a binary condition via a second command list plus a runtime assert) with a plain is_rack_only guard.
- Fix maasserver.eventloop.resolve_rpc_bind_addresses to reuse the shared, hardening-gated bind-resolution helper in provisioningserver.utils.network instead of a hand-rolled duplicate that always derived from maas_url regardless of hardening state.
- Close a FIPS audit-logging gap in the hmcz power driver: install a fips_tls_handshake hook on zhmcclient's requests-based HTTPS session (hmc/mscm/wedge already got equivalent treatment via ssh_utils; HMC Z's zhmcclient/requests transport had none).
- Extract a shared apply_fips_power_validation() helper for the power-params/password-complexity validation duplicated between AdminMachineForm.clean() and PodForm.clean().
- Move build_hardening_validation_kwargs() from maasservicelayer.services.hardening into maasserver.config (next to RegionConfiguration, which it reads): the service layer must not import maasserver, and RegionConfiguration's untyped ConfigurationOption descriptors broke pyright when referenced from a pyright-strict module.
- Allow the newly-introduced `cryptography.x509`/asymmetric-key imports in utilities/check-imports for maascommon and maasserver, following the existing "be selective about cryptography" pattern.
- Assorted dedup/cleanup: shared bind-validation-kwargs helper between start_up.py and config_hardening.py, a single _get_dbname() helper in dbupgrade.py, a shared _serialize_trusted_ssh_host_keys() in the power workflow, trimmed redundant comments/docstrings, removed duplicate tests, and consolidated repeated documentation.

### DIFF
--- a/src/maasserver/eventloop.py
+++ b/src/maasserver/eventloop.py
@@ -77,17 +77,12 @@ def make_RegionControllerService(postgresListener, dbtasks):
 def resolve_rpc_bind_addresses():
     """Return the addresses `RegionService`'s RPC listener should bind to.
 
-    An explicit ``rpc_bind`` list in regiond.conf wins verbatim. Otherwise,
-    derive a single address from ``maas_url`` instead of either extreme:
-    a hardcoded loopback default breaks rack connectivity out of the box
-    (rack controllers dial the region's real routable address regardless),
-    and a bare wildcard bind is avoidable in the common case where
-    ``maas_url`` already tells us which interface is reachable.
-
-    If ``maas_url`` itself cannot be resolved to a local address, this
-    falls back to loopback under hardening (a wildcard bind is not
-    allowed) or an empty list -- meaning "bind every interface" --
-    otherwise, mirroring `resolve_bind_address`.
+    Thin wrapper around `resolve_service_bind`, the same hardening-aware
+    bind-resolution helper every other region/rack service uses: an
+    explicit ``rpc_bind`` list in regiond.conf wins verbatim; otherwise,
+    under hardening, a single non-wildcard address is derived from
+    ``maas_url``; outside hardening, an empty list is returned, meaning
+    "bind every interface".
 
     :class:`~maasserver.ipc.IPCMasterService._getListenAddresses` calls
     this too, so whatever `RegionService` actually binds to is exactly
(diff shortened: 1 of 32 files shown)


src/maasserver/eventloop.py
```diff
@@ -77,40 +77,26 @@ def make_RegionControllerService(postgresListener, dbtasks):
...
-
-    If ``maas_url`` itself cannot be resolved to a local address, this
-    falls back to loopback under hardening (a wildcard bind is not
-    allowed) or an empty list -- meaning "bind every interface" --
-    otherwise, mirroring `resolve_bind_address`.
+    Thin wrapper around `resolve_service_bind`, the same hardening-aware
+    bind-resolution helper every other region/rack service uses: an
+    explicit ``rpc_bind`` list in regiond.conf wins verbatim; otherwise,
+    under hardening, a single non-wildcard address is derived from
+    ``maas_url``; outside hardening, an empty list is returned, meaning
```
tanzin8r, line 84: before, an unset `rpc_bind` still derived from `maas_url` even outside hardening. Now it binds all interfaces instead. PR description currently says no behavioral controls change. I believe that would count as one.

alexsander-souza: done

tanzin8r (approved):
Nice work! lgtm

PR #661, opened 2026-09-15
refactor: delete external-auth secret when disabling configauth

r00ta:
Previously, running `maas configauth` with all fields left blank kept the `external-auth` secret in the store and only cleared its values. This made the logic around checking if external auth is enabled ambigous, since the secret still existed with empty content. And when vault is configured, it required to retrieve the secret from vault and check its content. 

Delete the secret entirely when neither a Candid URL nor a RBAC URL is provided, so presence of the secret alone reliably indicates whether external authentication is configured. This is required by the upcoming removal of Candid/RBAC support, whose upgrade check relies on secret presence and, for Vault-backed deployments, cannot inspect secret values.


src/maasserver/management/commands/configauth.py
```diff
@@ -151,17 +151,24 @@ def get_auth_config(secret_manager):
...
+                "key": auth_details.key,
+                "admin-group": auth_details.admin_group,
+                "rbac-url": auth_details.rbac_url,
+            },
+        )
+    else:
+        # External authentication is being disabled: delete the secret
+        # entirely instead of storing empty values, so that its absence
+        # unambiguously means "not configured".
+        secret_manager.delete_secret("external-auth")
```
bjornt, line 171: Can this ever be called when there isn't any secret already there? for example, you configure local auth twice. If so, what happens? If it's idempotent, +1. Otherwise, do we need to add a check first?

bjornt (approved):
I added a question inline. But I'm also approving it, in case the answer is that it's already idempotent.

src/maasserver/management/commands/configauth.py
r00ta: yup, `delete_secret` is idempotent

PR #664, opened 2026-09-16
fix(storage): avoid naming collisions in get_available_name_for_node

skyecas:
Addresses the other edge case in https://bugs.launchpad.net/maas/+bug/2167420, withthe primary fixed by #512 

`get_available_name_for_node` only checked filesystemgroup names, and thus autogenerated names for bcache/virtual groups could collide

add an extra scan of blockdevice names to find the true next available index

Resolves LP:2167420


src/maasserver/models/filesystemgroup.py
```diff
@@ -136,7 +137,21 @@ def get_available_name_for_node(self, group_type, node):
...
         idx = -1
         for group in self.filter_by_node(node).filter(name__startswith=prefix):
-            name = group.name.replace(prefix, "")
+            name = group.name.removeprefix(prefix)
+            try:
+                name_idx = int(name)
+            except ValueError:
+                pass
+            else:
+                idx = max(idx, name_idx)
```
bjornt: This whole block is now duplicated in this method. I think it would be better to get all the names as strings that you want to consider, and then loop over those only once.

for name in fs_group_names + bd_names:
  ...

bjornt (changes requested):

bjornt (approved):

PR #667, opened 2026-09-16
chore: trigger LLM code review via /review PR comment

r00ta:
Add an `issue_comment` trigger so reviews can be requested on demand by commenting "/review" on a pull request, not just when the PR is opened.


.github/workflows/review-pr.yml
```diff
@@ -26,14 +32,35 @@ env:
 jobs:
   review:
     runs-on: ubuntu-latest
+    if: >
+      github.event_name == 'pull_request_target' ||
+      (github.event.issue.pull_request != null &&
+       startsWith(github.event.comment.body, '/review') &&
```
skatsaounis, line 38: nit: if we accept only `/review` why don't we use `github.event.comment.body == '/review'` for exacth matching?

skatsaounis (approved):
LGTM with a nit

.github/workflows/review-pr.yml
r00ta: It's just to accept comments like `/review ` or `/review\n` as valid triggers for the pipeline

r00ta: Btw we can also add into the comment additional instructions or model selection for the pipeline, something like `/review model=gemini-2.5-pro context=<more input context?>`

PR #678, opened 2026-09-21
fix: restore wildcard default for temporal_bind outside hardening

alexsander-souza:
resolve_bind_address() derived a bind address from maas_url whenever temporal_bind was unset, regardless of hardening_active. Outside hardening this diverged from every sibling bind key (rpc_bind, api_bind, ...), which only narrow off 0.0.0.0/:: under hardening, and from temporal's own pre-hardening default of binding 0.0.0.0 unconditionally.

In a VIP-fronted HA region cluster this meant Temporal bound to the node's own outbound-routing address instead of 0.0.0.0, so peers dialing the VIP on port 5271 got connection refused even though hardening was never enabled.

Gate the maas_url-derivation step on hardening_active, matching the resolve_bind_addresses() plural counterpart and the documented acceptance criteria for the hardening feature.


wyattrees (approved):
+1

PR #679, opened 2026-09-21
fix: change param key of DSN from ssl to sslmode

AloizioMacedo:
The param key name should be "sslmode", not "ssl".

This is invisible when using maas-test-db since that parameter is not added in that case.

Before this fix, this is the behavior we see:

```bash
MAAS URL [default=http://10.10.0.21:5240/MAAS]: 
Failed to perfom migrations:ations
Running OpenFGA migrations:
Failed to apply OpenFGA migrations
panic: failed to initialize database connection: failed to connect to `user=maas database=maasdb`:
        127.0.0.1:5432 (localhost): server error: FATAL: parameter "ssl" cannot be changed now (SQLSTATE 55P02)
        127.0.0.1:5432 (localhost): server error: FATAL: parameter "ssl" cannot be changed now (SQLSTATE 55P02)
```


wyattrees (approved):
+1

